PR #100, opened 2015-03-21
ENH: Use Welford's method for move_std

### LINKED ISSUE #99: Robustness of `move_std`
The algorithm used by `move_std` is prone to numerical errors. A while back I rewrote the implementation of the equivalent function in Pandas to use what some call [Welford's method](http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm), a one pass stable algorithm, see [here](https://github.com/pydata/pandas/pull/6817) for the relevant PR. I took some timings on the Pandas code, and it resulted in a 25% slowdown.

Would you like to see something like that implemented for bottleneck?

jaimefrio:
Closes #99

### DIFF
--- a/bottleneck/src/template/move.pyx
+++ b/bottleneck/src/template/move.pyx
@@ -382,53 +382,65 @@ cdef ndarray move_std_DTYPE0(ndarray a, int window, int min_count, int axis,
                              np.npy_intp* y_dims, int ddof):
     # bn.dtypes = [['float64'], ['float32']]
     cdef Py_ssize_t i, count
-    cdef DTYPE0_t asum, a2sum, ai, ssr, aold, yi
+    cdef DTYPE0_t delta, amean, assqdm, ai, aold, yi
     cdef ndarray y = PyArray_EMPTY(a_ndim, y_dims, NPY_DTYPE0, 0)
     cdef np.flatiter ity = PyArray_IterAllButAxis(y, &axis)
     cdef Py_ssize_t ystride = y.strides[axis]
     while PyArray_ITER_NOTDONE(ita):
-        asum = 0
-        a2sum = 0
+        amean = 0
+        assqdm = 0
         count = 0
         for i in range(min_count - 1):
             ai = (<DTYPE0_t*>((<char*>pid(ita)) + i*stride))[0]
             if ai == ai:
-                asum += ai
-                a2sum += ai * ai
                 count += 1
+                delta = ai - amean
+                amean += delta / count
+                assqdm += delta * (ai - amean)
             (<DTYPE0_t*>((<char*>pid(ity)) + i*ystride))[0] = NAN
         for i in range(min_count - 1, window):
             ai = (<DTYPE0_t*>((<char*>pid(ita)) + i*stride))[0]
             if ai == ai:
-                asum += ai
-                a2sum += ai * ai
                 count += 1
+                delta = ai - amean
+                amean += delta / count
+                assqdm += delta * (ai - amean)
             if count >= min_count:
-                ssr = a2sum - asum * asum / count
-                if ssr < 0:
-                    yi = 0
-                else:
-                    yi = sqrt(ssr / (count - ddof))
+                if assqdm < 0:
+                    assqdm = 0
+                yi = sqrt(assqdm / (count - ddof))
             else:
                 yi = NAN
             (<DTYPE0_t*>((<char*>pid(ity)) + i*ystride))[0] = yi
         for i in range(window, length):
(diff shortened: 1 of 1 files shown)


jaimefrio:
The slowdown is larger than I was hoping for the case with NaNs:

```
ottleneck performance benchmark
    Bottleneck  1.1.0dev
    Numpy (np)  1.10.0.dev0+36f742d
    NaN means approx one-third NaNs; float64 and axis=-1 are used

                 no NaN     no NaN      NaN        NaN    
                   (10,)   (1000,1000)   (10,)   (1000,1000)
    move_std      965.0      111.5     1014.5      544.1  <-- master
    move_std      957.4       87.7     1002.3      230.9  <-- this PR
```

But the numerical stability is greatly improved:

```
# master
>>> a = np.random.rand(10)
>>> bn.move_std(a, 3)
array([        nan,         nan,  0.35305001,  0.08837788,  0.22345456,
        0.15607194,  0.16602772,  0.20010568,  0.28201036,  0.37055974])
>>> bn.move_std(a+1e6, 3)
array([        nan,         nan,  0.35286218,  0.08838835,  0.223534  ,
        0.15676997,  0.16634083,  0.20050392,  0.28269306,  0.37085381])
>>> bn.move_std(a+1e7, 3)
array([        nan,         nan,  0.32274861,  0.        ,  0.20412415,
        0.14433757,  0.25      ,  0.14433757,  0.25      ,  0.35355339])
>>> bn.move_std(a+1e8, 3)
array([        nan,         nan,  0.        ,  0.        ,  0.        ,
        0.        ,  0.        ,  2.        ,  2.        ,  2.30940108])

# This PR
>>> a = np.random.rand(10)
>>> bn.move_std(a, 3)
array([        nan,         nan,  0.25058293,  0.2550371 ,  0.26414995,
        0.28272484,  0.29881878,  0.29482359,  0.237722  ,  0.19583085])
>>> bn.move_std(a+1e6, 3)
array([        nan,         nan,  0.25058293,  0.2550371 ,  0.26414995,
        0.28272484,  0.29881878,  0.29482359,  0.237722  ,  0.19583085])
>>> bn.move_std(a+1e9, 3)
array([        nan,         nan,  0.25058295,  0.25503715,  0.26415001,
        0.28272492,  0.29881885,  0.29482366,  0.23772199,  0.19583086])
>>> bn.move_std(a+1e12, 3)
array([        nan,         nan,  0.25054534,  0.25502679,  0.26414348,
        0.28270348,  0.2987971 ,  0.29479017,  0.23766572,  0.19576629])
>>> bn.move_std(a+1e15, 3)
array([        nan,         nan,  0.22821773,  0.22821773,  0.22821773,
        0.26020825,  0.26020825,  0.26020825,  0.21650635,  0.1767767 ])
>>> bn.move_std(a+1e18, 3)
array([ nan,  nan,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.])
```

kwgoodman:
Here are the timings on my computer:

```
    Bottleneck  1.1.0dev
    Numpy (np)  1.9.1
    NaN means approx one-third NaNs; float64 and axis=-1 are used

                 no NaN     no NaN      NaN        NaN    
                   (10,)   (1000,1000)   (10,)   (1000,1000)
    move_std     1361.3       56.2     1483.4      762.4   master
    move_std     1379.7       52.9     1488.4      299.8   pr
```

Some of our choices:

A. Leave ``move_std`` as is; add warning about large means to docstring
B. Use slower, more robust algo
C. Keep both algos, changing name of current algo

kwgoodman:
The second ``move_std`` in the template handles int input arrays. (I know, it's confusing.) So there is no need for the ``if ai == ai`` check in the second ``move_std``.

jaimefrio:
I sort of figured that out, but only implemented it for the bulk of the array, not for the first few items. Will change.

kwgoodman:
I think we should go with option B. @shoyer?

What's left to be done on this PR? I'm sure we'd come up with the same list. Here's mine.

Adding the following lines to the end of the arrays function in bottleneck/tests/move_test.py:

    # check that move_std is robust
    yield np.array([1., 2., 3.]) + 1e9

fails in master but passes in this PR.

Rewrite docstring.

Add a **Beware** section to RELEASE.rst. Something like move_std is slower but more robust.

I think this is a nice improvement to bottleneck. Thank you.

kwgoodman:
Oh, I'll update the C file and the benchmark in the readme.

shoyer:
I agree -- way better to be robust (option B) than to be fast and inaccurate. With bottleneck, rolling window calculations aren't really a bottleneck anymore, so a factor of 3 slow down would not be the end of the world

PR #107, opened 2016-01-09
Minimal change in metadata that fixes the top_level.txt file

### LINKED ISSUE #106: Broken top_level.txt file in distribution 1.0.0
This is what the [setuptools documentation](https://pythonhosted.org/setuptools/formats.html#top-level-txt-conflict-management-metadata) says about the `top_level.txt`file:

> This file is a list of the top-level module or package names provided by the project, one Python identifier per line.
>
> Subpackages are not included; a project containing both a ``foo.bar`` and a ``foo.baz`` would include only one line, ``foo``, in its ``top_level.txt``.

The file distributed with Bottleneck 1.0.0 can be accessed with:

    >>> import pkg_resources
    >>> provider = pkg_resources.get_distribution('Bottleneck')
    >>> print(provider.get_metadata('top_level.txt'))
    bottleneck
    bottleneck/benchmark
    bottleneck/slow
    bottleneck/src
    bottleneck/src/auto_pyx
    bottleneck/src/template
    bottleneck/tests
    move
    nonreduce
    nonreduce_axis
    reduce

and is clearly broken.

I'm not sure, but I think the problem is that the `PACKAGES` variable in `setup.py` is not what should be passed to the `packages=` keyword argument.

alexamici:
This fixes #106 and looks ok to my non-expert eyes. The fix has two components:
- The paths in `PACKAGES` were certainly wrong, I could use `find_packages()`, but the explicit `'bottleneck'` is ok as well.
- I'm not sure if you have a reason for using `ext_package` instead of full names in `ext_modules`, but the looking at the [python distribution docs](https://docs.python.org/2/distutils/setupscript.html#extension-names-and-packages) it feels safe.

Tests pass on my machine for 2.7 and 3.5, except the known failure in the median.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,13 +57,7 @@ MINOR = VER[1]
 MICRO = VER[2]
 ISRELEASED = False
 VERSION = '%s.%s.%s' % (MAJOR, MINOR, MICRO)
-PACKAGES = ["bottleneck",
-            "bottleneck/slow",
-            "bottleneck/tests",
-            "bottleneck/benchmark",
-            "bottleneck/src",
-            "bottleneck/src/template",
-            "bottleneck/src/auto_pyx"]
+PACKAGES = ['bottleneck']
 PACKAGE_DATA = {'bottleneck': ['LICENSE']}
 REQUIRES = ["numpy"]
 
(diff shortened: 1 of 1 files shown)


kwgoodman:
I like that your change makes the file simpler and easier to maintain.

What version of numpy are you using that gives you a unit test failure with median?

I'll give your change a try when I get a chance. If anyone else tries it let's us know how it went.

alexamici:
`numpy==1.10.1`

This my full test session with python 2:
```
$ ../venv/bin/python
Python 2.7.10 (default, Jul 13 2015, 12:05:58) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import bottleneck
>>> bottleneck.test()
Running unit tests for bottleneck
NumPy version 1.10.1
NumPy relaxed strides checking option: True
NumPy is installed in /Users/amici/devel/mac-cpython2/lib/python2.7/site-packages/numpy
Python version 2.7.10 (default, Jul 13 2015, 12:05:58) [GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]
nose version 1.3.7
...................................................................F..........
======================================================================
FAIL: Test median.
----------------------------------------------------------------------
  File "/Users/amici/devel/mac-cpython2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "bottleneck/tests/reduce_test.py", line 95, in unit_maker
    assert_array_equal(actual, desired, err_msg)
  File "/Users/amici/devel/mac-cpython2/lib/python2.7/site-packages/numpy/testing/utils.py", line 782, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "/Users/amici/devel/mac-cpython2/lib/python2.7/site-packages/numpy/testing/utils.py", line 708, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

func median | input a0 (float64) | shape (0,) | axis -1

Input array:
[]

(mismatch 100.0%)
 x: array(nan)
 y: array('Crashed', 
      dtype='|S7')

----------------------------------------------------------------------
Ran 79 tests in 82.754s

FAILED (failures=1)
<nose.result.TextTestResult run=79 errors=0 failures=1>

```

alexamici:
Hang on, tox fails with a suspect:

    ImportError: cannot import name slow

for python 2.7. I'm investigating.

PR #110, opened 2016-02-23
move_median now works with nans, and added a min_count parameter

jennolsen84:


kwgoodman:
Sorry to be so late to respond, especially so since this is such great work.

Some quick comments/suggestions:

- Add a note to the RELEASE.rst file
- A couple of the functions in move_median.c are commented out (check_asserts and mm_dump). Uncomment and either remove calls to these functions or comment them out
- For integer input arrays we can call mm_update_movemedian_nonan in function move_median_DTYPE0 in file move.pyx. I assume that is faster. Perhaps we can split move_median_DTYPE0 into two functions (floats, ints) as is done for the other functions such as move_sum
- You mentioned writing a description of the algo. Because I don't kow how it works I think that is a good idea. If you are up for writing a detailed description, perhaps including a small explicit example such as a 6 element input array with something like a window of 3, then you could add an Algorithms section to the sphynx doc. But that would be a lot of work 
- Update docstring

kwgoodman:
@jennolsen84 I went ahead and merged this after taking care of the suggestions I made above. Did I miss any important details?

Issue #114, opened 2016-05-24
Debian and build system


toobaz:
I would like to create a Debian package for bottleneck.

Since (according to the Debian policy) I need to regenerate the ``*.c`` files in the build process (rather than using the ones shipped in the tarball), and since I want to deviate as least as possible from the upstream build process, it is my understanding that I should run first each ``*_setup.py`` inside ``bottleneck/scr/auto_pyx`` (and then the main ``setup.py``). However:

```
~/bottleneck-1.0.0$ cd bottleneck/src/auto_pyx/
~/bottleneck-1.0.0/bottleneck/src/auto_pyx$ python\
nonreduce_setup.py build_ext --inplace
running build_ext
cythoning nonreduce.pyx to nonreduce.c
error: [Errno 2] No such file or directory: '/home/pietro/bottleneck-
1.0.0/bottleneck/src/auto_pyx/nonreduce.pyx'
```

Indeed, the ``*.pyx`` files are in ``templates``. What is the intended use of the ``*_setup.py`` files?

kwgoodman:
Getting Bottleneck into Debian would be great!

The auto-generated C files are a pain in the ass. They are large which makes `git diff` painful and bloats the size of the tarball. And they are a security risk because someone (me) could slip in some code during a pull request. Also if someone commits a change to a pyx file without updating the C files then our newly set up CI, which runs off of the C files, will not unit test the change.

So I am interested in pulling out the auto-generated C files and instead have the top-level setup.py construct them at install time. I think that's what pandas does. If anyone is interested in working on that, let me know. The downside, of course, is that installing bottleneck would now require cython. And that I only test with one version of cython.

I never use the auto-gen C files in the repo. I always regenerate with `make all`. So to answer your question on how to generate the C files with the current release (1.0), take a look at the top-level makefile.

Let me know if you run into any problems.

toobaz:
Yes, pandas (and most other packages I can think of) generates the *.c files in the top-level ``setup.py``. This can actually be [very simple to do](http://pietrobattiston.it/gitweb?p=jsaone.git;a=blob;f=setup.py;h=777a484e37999ceaaa25e5718cadff5393bfa095;hb=HEAD#l17).

By the way, pandas does ship C files in the releases, but does not keep them in git - they are generated at release time. And it's also simple to write a ``setup.py`` which will generate them if Cython is installed and otherwise use the shipped ones.

[...]

PR #115, opened 2016-05-24
Single setup py

toobaz:
OK, this is related to #114 .

I was a bit drastic... I won't get offended if you want to reject something ;-)

Anyway:
```
python setup.py build_ext --inplace
python3 setup.py build_ext --inplace
```
will now generate the automatically generated .pyx _and_ the .c files _if Cython is installed_. If it's not, they will fallback to the existing .c files (which are assumed to be shipped in the release tarball - according with ``MANIFEST.in``). I ran the tests and everything seems to work fine.

By the way,
```
python setup.py build_sphinx
```

will generate the documentation. This does not work with python3 because of a SyntaxError I didn't bother to fix (fixing it is trivial, keeping compatibility with both python 2 and 3 is not, at least for me).


kwgoodman:
The PR is failing on travis (probably because our travis.yml file does not install cython and the auto-gen c files are not in the repo anymore [which is great!]): https://travis-ci.org/kwgoodman/bottleneck/builds/132676230

The top-level makefile got deleted. I need that one. The doc makefile too. Good catch on the py3 compat of the docs. I guess we should add a unit test that makes the docs (we can keep that separate from this PR---you've got enough on your hands).

kwgoodman:
Oh, sorry, nevermind about the makefiles. Very nice that you got rid of them. I'll add back a smaller top-level makefile later.

toobaz:
Concerning the makefiles: I told you I was a bit drastic! ;-) Feel free to filter my changes.

Concerning Travis, yes, your diagnosis is correct. I never played with Travis however. I can try to add a ``conda install cython`` at the end of ``ci/travis/conda_setup.sh``...

kwgoodman:
@Midnighter to add cython to travis do we just add cython to
```
env: DEPS="numpy==1.11.0 nose" PYTHON_VERSION="2.7" PYTHON_ARCH="64"
```
in .tavis.yml?

Midnighter:
Exactly, just put all required packages in the `DEPS="..."` string and conda will install them when the environment is generated.

Btw, I'm a big fan of the following [cookiecutter](https://github.com/ionelmc/cookiecutter-pylibrary) that generates a library project skeleton for you. There the docs are also only generated with Python 2.7. Seems to be a general thing.

toobaz:
Uhm... now I'm clueless.

Midnighter:
For extensions the option is called `include_dirs=np.get_include()` is that it instead of `include_path`?

kwgoodman:
It is now passing on linux but not osx. On osx the moving window functions don't get built. Unfortunately travis [does not show](https://travis-ci.org/kwgoodman/bottleneck/jobs/132884953#L248) the compiling of bottleneck. Is there anyway to get travis to show it? Should look something like:

```
kg@kg:/devel/bottleneck (unroll *)$ ma
rm -rf build dist Bottleneck.egg-info
find . -name \*.pyc -delete
rm -rf bottleneck/src/auto_pyx/*.c bottleneck/src/auto_pyx/*.html bottleneck/src/auto_pyx/build bottleneck/src/auto_pyx/../../*.so
python -c "from bottleneck.src.template.template import make_pyx; make_pyx();"
cython bottleneck/src/auto_pyx/reduce.pyx
cython bottleneck/src/auto_pyx/nonreduce.pyx
cython bottleneck/src/auto_pyx/nonreduce_axis.pyx
cython bottleneck/src/auto_pyx/move.pyx
rm -rf bottleneck/src/auto_pyx/../reduce.so
python bottleneck/src/auto_pyx/reduce_setup.py build_ext --inplace
running build_ext
skipping 'bottleneck/src/auto_pyx/reduce.c' Cython extension (up-to-date)
building 'reduce' extension
creating build
creating build/temp.linux-x86_64-2.7
creating build/temp.linux-x86_64-2.7/bottleneck
creating build/temp.linux-x86_64-2.7/bottleneck/src
creating build/temp.linux-x86_64-2.7/bottleneck/src/auto_pyx
x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/local/lib/python2.7/dist-packages/numpy/core/include -I/usr/include/python2.7 -c bottleneck/src/auto_pyx/reduce.c -o build/temp.linux-x86_64-2.7/bottleneck/src/auto_pyx/reduce.o
In file included from /usr/local/lib/python2.7/dist-packages/numpy/core/include/numpy/ndarraytypes.h:1777:0,
                 from /usr/local/lib/python2.7/dist-packages/numpy/core/include/numpy/ndarrayobject.h:18,
                 from /usr/local/lib/python2.7/dist-packages/numpy/core/include/numpy/arrayobject.h:4,
                 from bottleneck/src/auto_pyx/reduce.c:266:
<snip>
```

That should show us the error which would be nice since I don't have osx.

toobaz:
Although everything worked here, I kept in me a sense of guilt for having removed that flag! Now I feel better.

kwgoodman:
Looks good. Could you make some tweaks?

- put `bottleneck/template/__init__.py` back into `bottleneck/template/template.py`
- revive the top-level makefile, updating it to use the setup.py file (which will cut out a lot of junk)

kwgoodman:
tox no longer runs:
```
kg@kg:/devel/bottleneck (toobaz-single_setup_py)$ tox
<snip>
bottleneck/move.c:287:30: fatal error: csrc/move_median.c: No such file or directory
 #include "csrc/move_median.c"
<snip>
```

It is because the csrc directory doesn't appear in the sdist.

Midnighter:
> On osx the moving window functions don't get built. Unfortunately travis does not show the compiling of bottleneck. Is there anyway to get travis to show it?

At the moment the `.travis.yml` contains the line `pip install .` which swallows that output. If you change that to `python setup.py install` instead (or `build_ext`) the output should be there.

.travis.yml
```diff
@@ -3,17 +3,17 @@ sudo: false
 matrix:
     include:
         - os: linux
-          env: DEPS="numpy==1.11.0 nose" PYTHON_VERSION="2.7" PYTHON_ARCH="64"
+          env: DEPS="numpy==1.11.0 nose cython" PYTHON_VERSION="2.7" PYTHON_ARCH="64"
```
Midnighter: Since you're pinning the numpy version it might be a good idea to do that for Cython as well.

kwgoodman:
On travis we are currently only testing installation on a system that has cython installed. Would be nice to also test without cython. To do so I guess we'd have to make an sdist and then install that. Not sure how to do that. Point pip at the sdist archive might work.

That would double our number of environments to 12. Maybe we could drop python 3.4? Which would give us 8.

We can do this is a separate PR.

Midnighter:
For the separate pull request:

1. I didn't find it earlier but `pip` has a `--verbose` option which gives you all the output that you wanted to see.
2. Running `python setup.py sdist` creates an archive like this: `dist/Bottleneck-1.1.0.dev0.tar.gz` which can then be installed with `pip`. I'd be happy to make a pull request that adapts those changes in the `.travis.yml` once you're satisfied with the state of the master branch.

PR #119, opened 2016-05-26
Add Travis-CI status to README.rst

shoyer:
This where people usually look to find the CI status.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,9 @@
 Bottleneck
 ==========
 
+.. image:: https://travis-ci.org/kwgoodman/bottleneck.svg?branch=master
+    :target: https://travis-ci.org/kwgoodman/bottleneck
+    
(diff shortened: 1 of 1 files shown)


kwgoodman:
Thank you. I've been meaning to do that. The reason I put it off is that the readme file is part of the sphinx doc. See https://github.com/kwgoodman/bottleneck/blob/master/doc/source/intro.rst

So we'd have to somehow filter out the travis status when building the docs. I'm unsure of how to make that happen. Putting the status at the very top of the file will make things easier.

shoyer:
Ah, that's a little awkward. I'll close this for now while you figure out the docs thing.

kwgoodman:
I added the travis status: https://github.com/kwgoodman/bottleneck/commit/398ede36f61b7ae4786286defedaa2c17cf12869

PR #120, opened 2016-05-27
Expand testing to sdist and flake8

Midnighter:
I left in Python 3.4 since it's the standard version for Debian and Ubuntu (which I'm using currently). Not sure about the intricacies of differences between 3.4 and 3.5 but seems that there's no harm done. Everything runs fine except for flake8 which fails on some of the test files. You can either change those files or adjust the way that flake8 runs (by excluding test files in `setup.cfg`, for example).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,31 @@ sudo: false
 matrix:
     include:
         - os: linux
+          env: DEPS="flake8" PYTHON_VERSION="2.7" PYTHON_ARCH="64" RUN="style"
+        - os: linux
           env: DEPS="numpy==1.11.0 nose cython" PYTHON_VERSION="2.7" PYTHON_ARCH="64"
         - os: linux
           env: DEPS="numpy==1.11.0 nose cython" PYTHON_VERSION="3.4" PYTHON_ARCH="64"
         - os: linux
           env: DEPS="numpy==1.11.0 nose cython" PYTHON_VERSION="3.5" PYTHON_ARCH="64"
+        - os: linux
+          env: DEPS="numpy==1.11.0 nose cython" PYTHON_VERSION="2.7" PYTHON_ARCH="64" RUN="sdist"
+        - os: linux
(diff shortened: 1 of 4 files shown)


kwgoodman:
Flake8 is a nice addition. I upgraded flake8 on my computer and now see the same failures. I can deal with those after this PR goes in.

It confused me to see cython installed in `.travis.yml` and then uninstalled in `bn_setup.sh` until I realized that cython is now required (since #115) to make the sdist

I think we can do this in 6 tests (not including flake8) instead of 12. With limited concurrency on travis more tests means longer run times. Does this sound reasonable:

linux:
- py27, cython, no sdist
- py34, cython, sdist
- py35, no cython, sdist

osx:
- py27, no cython, sdist
- py34, cython, no sdist
- py35, cython, sdist

Midnighter:
Limiting the number of tests is surely possible. I'll also play with caching the download of conda to speed up tests. I'm not sure what you mean by "no cython, sdist", how is that supposed to run since cython is needed to create the sdist?

kwgoodman:
Build 25 for this PR took 27 min 38 sec if I am reading it right. I had some builds timeout yesterday waiting to download stuff from continuum (e.g. build 21.2). I guess more and more people are using miniconda.

Caching sounds interesting.

Yeah I made that confusing.  By cython and no cython I just meant at bn install time. So with no cython we'd uninstall cython after making the sdist.

Midnighter:
> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source [docs](https://docs.travis-ci.com/user/caching/#Things-not-to-cache)

I'll leave caching for some other time then.

kwgoodman:
Should we test the case of installing from a sdist when cython is installed?

Midnighter:
I think this is what you asked for. I squashed a few commits, so they disappeared here.

kwgoodman:
This PR is a nice improvement to bottleneck. I'll merge.

PR #129, opened 2016-06-20
Fix compile errors with msvc9 and msvc10

cgohlke:
I took the [definition of NAN from Cython](https://github.com/cython/cython/blob/master/Cython/Utility/ModuleSetupCode.c#L309).
See also https://github.com/kwgoodman/bottleneck/issues/128

### DIFF
--- a/bottleneck/csrc/move_median.h
+++ b/bottleneck/csrc/move_median.h
@@ -4,6 +4,16 @@
 #include <math.h>
 #include <assert.h>
 
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+#define inline __inline
+static __inline float __NAN() {
+  float value;
+  memset(&value, 0xFF, sizeof(value));
+  return value;
+}
+#define NAN __NAN()
+#endif
+
 #if NOCYTHON==1
 
     typedef size_t idx_t;
--- a/bottleneck/csrc/move_median.c
+++ b/bottleneck/csrc/move_median.c
@@ -265,6 +265,12 @@ mm_update_init_nan(mm_handle *mm, ai_t ai)
 inline ai_t
 mm_update_nan(mm_handle *mm, ai_t ai)
 {
+    idx_t n_s, n_l, n_n;
+
+    mm_node **l_heap;
+    mm_node **s_heap;
+    mm_node **n_array;
+    mm_node *node2;
 
     // node is oldest node with ai of newest node
     mm_node *node = mm->oldest;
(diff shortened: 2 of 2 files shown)


bottleneck/csrc/move_median.h
```diff
@@ -4,6 +4,16 @@
...
 #include <assert.h>
 
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+#define inline __inline
+static __inline float __NAN() {
+  float value;
+  memset(&value, 0xFF, sizeof(value));
+  return value;
+}
+#define NAN __NAN()
```
jennolsen84: @kwgoodman what if we changed line 14 to:

```
const double NAN = __NAN()
const double NAN64 = __NAN()
const float NAN32 = __NAN()
```

This will avoid calling memset() repeatedly, as there will be only one NAN value.  Also, later down the line, we can turn move_median into a template so that it is fast for float32 as well.

kwgoodman: @jennolsen84, I had trouble redefining NAN using `const double NAN = __NAN()`. I ended up trying:
```
kg@kg:/devel/bottleneck (cgohlke-patch-1 *)$ di
diff --git a/bottleneck/csrc/move_median.c b/bottleneck/csrc/move_median.c
index cf94d83..ff5d477 100644
--- a/bottleneck/csrc/move_median.c
+++ b/bottleneck/csrc/move_median.c
@@ -10,6 +10,11 @@ node1->idx = idx2;                                 \
 node2->idx = idx1;                                 \
 idx1       = idx2
 
+/* define nan */
+ai_t myNAN;
+void init_nan(void) {
+  memset(&myNAN, 0xFF, sizeof(ai_t));
+}
 
 /*
 -----------------------------------------------------------------------------
@@ -64,6 +69,8 @@ mm_new(const idx_t window, idx_t min_count)
 
     mm_reset(mm);
 
+    init_nan();
+
     return mm;
 }
 
@@ -511,7 +518,7 @@ mm_get_median(mm_handle *mm)
 {
     idx_t n_total = mm->n_l + mm->n_s;
     if (n_total < mm->min_count)
-        return NAN;
+        return myNAN;
     if (min(mm->window, n_total) % 2 == 1)
         return mm->s_heap[0]->ai;
     return (mm->s_heap[0]->ai + mm->l_heap[0]->ai) / 2.0;
diff --git a/bottleneck/csrc/move_median.h b/bottleneck/csrc/move_median.h
index d6bbe54..736b0b0 100644
--- a/bottleneck/csrc/move_median.h
+++ b/bottleneck/csrc/move_median.h
@@ -6,12 +6,6 @@
 
 #if defined(_MSC_VER) && (_MSC_VER < 1900)
 #define inline __inline
-static __inline float __NAN() {
-  float value;
-  memset(&value, 0xFF, sizeof(value));
-  return value;
-}
-#define NAN __NAN()
 #endif
 
 #if NOCYTHON==1
```

Note that NAN is only used in one place---calculating the median when the count is less than `min_count`.  These changes slow the code a bit, but I wasn't too careful in my measurement.

PR #170, opened 2017-06-04
Enable build of docs with Python 3

ghisvail:
`file()` is no longer a built-in in Python 3. It is better to just use a single call to `open` chained with a `read`.


kwgoodman:
Thank you. I still use py 2.7 so I didn't notice. I guess we should build the docs on travis to at least see that they run.

PR #199, opened 2018-12-12
Silence NumPy warnings about indexing with non-tuples

shoyer:
Fixes https://github.com/kwgoodman/bottleneck/issues/194

These show up when using the "slow" versions of moving window functions, e.g.,

  FutureWarning: Using a non-tuple sequence for multidimensional indexing is
  deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this
  will be interpreted as an array index, `arr[np.array(seq)]`, which will
  result either in an error or a different result.
    nidx1 = n[idx1]

Let me know if you'd like me to add a test here (but I think these lines already have coverage).


kwgoodman:
Awesome. Thank you. Could you add a line to RELEASE.rst?

shoyer:
Done!

Issue #158, opened 2016-11-01
Error installing Bottleneck 1.2.0 from behind proxy


gfairchild:
I just ran into an error where some unit tests are failing after upgrading from 1.1.0 to 1.2.0. This is on a work-hosted GitLab instance that's behind a proxy. Here's my virtual environment:

```
amqp==1.4.9
anyjson==0.3.3
autopep8==1.2.4
billiard==3.5.0.1
Bottleneck==1.2.0
celery==3.1.24
colormath==2.1.1
decorator==4.0.10
Django==1.9.11
django-celery==3.1.17
django-debug-toolbar==1.5
django-extensions==1.7.4
django-filter==0.15.3
django-nested-admin==3.0.12
django-widget-tweaks==1.4.1
djangorestframework==3.5.2
factory-boy==2.7.0
fake-factory==0.7.2
isodate==0.5.4
isort==4.2.5
kombu==3.0.35
Markdown==2.6.7
networkx==1.11
numexpr==2.6.1
numpy==1.11.2
pandas==0.19.0
pep8==1.7.0
psycopg2==2.6.2
py==1.4.31
pyflakes==1.3.0
pyHarmonySearch==1.3.3
pylibmc==1.5.1
pyshp==1.2.10
pytest==3.0.3
pytest-django==3.0.0
python-dateutil==2.5.3
python-monkey-business==1.0.0
pytz==2016.7
requests==2.11.1
scipy==0.18.1
six==1.10.0
spectra==0.0.7
sqlparse==0.2.2
vine==1.1.3
whitenoise==3.2.2
xlrd==1.0.0
```

Here's the output of my CI output attempting to install 1.2.0:

```
...
...
...
$ python3 -m venv env
$ source env/bin/activate
$ pip --proxy $HTTP_PROXY install --upgrade pip
Collecting pip
  Downloading pip-8.1.2-py2.py3-none-any.whl (1.2MB)
Installing collected packages: pip
  Found existing installation: pip 8.1.1
    Uninstalling pip-8.1.1:
      Successfully uninstalled pip-8.1.1
Successfully installed pip-8.1.2
$ pip --proxy $HTTP_PROXY install -r requirements.txt
Collecting amqp==1.4.9 (from -r requirements.txt (line 1))
  Downloading amqp-1.4.9-py2.py3-none-any.whl (51kB)
Collecting anyjson==0.3.3 (from -r requirements.txt (line 2))
  Downloading anyjson-0.3.3.tar.gz
[...]

PR #203, opened 2019-01-07
Add a pyproject.toml file declaring build-time dependencies

astrofrog:
The use of ``setup_requires`` to declare build-time dependencies is problematic because it is honored by setuptools rather than pip, which means that it doesn't ignore numpy pre-releases. This in turn is causing issues at the moment because when installing bottleneck into a clean environment, it is build against numpy 1.16rc2 and then numpy 1.15 is installed, which then causes bottleneck to not work (since there was a change in API between numpy 1.15 and 1.16). For a more in-depth description of this issue, see [this post I made on numpy-discussion](https://mail.python.org/pipermail/numpy-discussion/2019-January/079097.html).

In any case, projects should now start declaring build-time dependencies using a ``pyproject.toml`` file as described in [PEP 518](https://www.python.org/dev/peps/pep-0518/). Using this (as done in this PR) will mean that this issue won't happen for pip 10.0 and later since pip will honor the build-time dependency and ``setup_requires`` will thus get ignored.

In the long term, ``setup_requires`` could in fact be removed, but there's no harm in leaving it for older versions of pip.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 include README.rst RELEASE.rst
 include LICENSE bottleneck/LICENSE
 include Makefile tox.ini
+include pyproject.toml
 
 recursive-include bottleneck/src *.c *.h
 exclude bottleneck/src/reduce.c
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel", "numpy"]


kwgoodman:
Seems like a lot of people have been hitting this problem so a solution is great!

Can anyone confirm that this PR works or doesn't work for them? @drdavella can you confirm that this PR fixes your issue as well?

kwgoodman:
@gfairchild I wonder if this PR also fixes your issue (#158)

gfairchild:
I was just reading PEP 518 and was wondering the same! Gonna try it now.

astrofrog:
For anyone who wants to try this, the following should work:

```
pip install git+https://github.com/astrofrog/bottleneck.git@pep-518
```

kwgoodman:
@astrofrog I've never used the wheel package. Could you comment on whether that is required in the PR or just useful.

astrofrog:
@kwgoodman - the setuptools and wheel entries are just what is the default defined in PEP518 if no ``pyproject.toml`` is found, so I thought it'd be safest to leave it and just add numpy. When pip installs packages, it internally builds wheels before installing them.

Here's an example taken from pip log when installing bottleneck:

```
Building wheels for collected packages: bottleneck
  Running setup.py bdist_wheel for bottleneck ... done
  Stored in directory: /root/.cache/pip/wheels/f2/bf/ec/e0f39aa27001525ad455139ee57ec7d0776fe074dfd78c97e4
Successfully built bottleneck
```

kwgoodman:
@astrofrog That sounds good.

kwgoodman:
I installed bottleneck on a system that already had numpy installed using

```sudo pip install git+https://github.com/astrofrog/bottleneck.git@pep-518```

it worked.

astrofrog:
So one thing that won't work properly with this PR: ``pip install bottleneck numpy==1.13.1`` - this is because then bottleneck will first be compiled with Numpy 1.15 and then 1.13 will be installed. For things to work properly we actually need to pin numpy to the latest compatible version. However, this is dependent on Python version - see https://github.com/astrofrog/fast-histogram/issues/35 and https://github.com/scipy/scipy/issues/8734 for more details. Unfortunately it seems like this requires pip 18+ which is the current version.

Issue #191, opened 2018-05-30
Preparing to release bottleneck 1.3.0


kwgoodman:
I am getting ready to release bottleneck 1.2.2. The only thing left to do is testing.

The following people gave test reports on the pre-release of bottleneck 1.2.1, so I'm pinging you again in case you have time to test this release (the master branch): @cgohlke @itdaniher @toobaz @shoyer and anyone else.

shoyer:
Xarray's CI is passing with the development version of bottleneck, so I don't anticipate any troubles here:
https://travis-ci.org/pydata/xarray/jobs/385018063

cgohlke:
All builds and test are passing on Windows with numpy 1.14.3

V-Kh:
I've tried to build an Arch package and I get this: 

`bottleneck/src/move.c: In function ‘move_rank_int64’:
bottleneck/src/move.c:2009:24: warning: self-comparison always evaluates to true [-Wtautological-compare]
                 if (aj == aj) {
                        ^~
bottleneck/src/move.c: In function ‘move_rank_int32’:
bottleneck/src/move.c:2070:24: warning: self-comparison always evaluates to true [-Wtautological-compare]
                 if (aj == aj) {`

It also seems that my makepkg options conflict with your default compiler flags because I see such cryptic messages:
`gcc -pthread -DNDEBUG -g -fwrapv -O3 -Wall -march=native -O2 -pipe -fstack-protector-strong -fno-plt -D_FORTIFY_SOURCE=2 -fPIC -I/usr/lib/python3.7/site-packages/numpy/core/include -I/usr/include/python3.7m -c bottleneck/src/reduce.c -o build/temp.linux-x86_64-3.7/bottleneck/src/reduce.o -O2
gcc -pthread -shared -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -march=native -O2 -pipe -fstack-protector-strong -fno-plt -D_FORTIFY_SOURCE=2 build/temp.linux-x86_64-3.7/bottleneck/src/reduce.o -L/usr/lib -lpython3.7m -o build/lib.linux-x86_64-3.7/bottleneck/reduce.cpython-37m-x86_64-linux-gnu.so
`

kwgoodman:
[...]

PR #206, opened 2019-04-26
Fix build error `AttributeError: 'dict' object has no attribute ...`

ales-erjavec:
Running `pip wheel ./`  in the source checkout with pip>19.0 fails with a `AttributeError: 'dict' object has no attribute __NUMPY_SETUP__`

Depending on how the setup.py script is executed (as `__main__` or not) `__builtins__` in the global namespace can be a dictionary or a module (https://docs.python.org/3/library/builtins.html says it can be either and that this is an cpython implementation detail).

Fixed by an explicit import.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,11 @@ class build_ext(_build_ext):
     def finalize_options(self):
         _build_ext.finalize_options(self)
         # prevent numpy from thinking it is still in its setup process
-        __builtins__.__NUMPY_SETUP__ = False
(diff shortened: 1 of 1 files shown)


kwgoodman:
Hey thanks! Right now master is failing unit tests due to #203 and as mentioned in #191 so I'll wait for those to pass before merging PRs.

qwhelan:
@ales-erjavec Thanks, merging!

PR #234, opened 2019-09-15
BLD: Add coverage

qwhelan:


qwhelan:
Closing as codecov can only report coverage of checked in files

PR #271, opened 2019-11-09
Do not include bn_config.h in source distribution

cgohlke:
Fix compile errors when building Bottleneck-1.3.0rc1 with msvc on Windows using the source distribution on PyPI:
```
bottleneck/src/reduce.c(2388): error C2143: syntax error: missing ')' before '('
bottleneck/src/reduce.c(2388): error C2091: function returns function
bottleneck/src/reduce.c(2388): error C2143: syntax error: missing ')' before 'string'
bottleneck/src/reduce.c(2388): error C2143: syntax error: missing '{' before 'string'
bottleneck/src/reduce.c(2388): error C2059: syntax error: 'string'
...
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,6 +10,7 @@ exclude bottleneck/src/reduce.c
 exclude bottleneck/src/move.c
 exclude bottleneck/src/nonreduce.c
 exclude bottleneck/src/nonreduce_axis.c
+exclude bottleneck/src/bn_config.h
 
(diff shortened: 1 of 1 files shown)


qwhelan:
Thanks for catching this! Slipped through as every CI test is using a package it builds itself, so this sort of cross-platform bug won't surface. Created #272 to solve this going forward.

I'll publish a `rc2` with this diff in an hour or so.

Issue #277, opened 2019-11-16
Installation with pip fails (on Arch Linux)


goggle:
I wanted to setup a development build for pandas, but this fails because of an installation error in `bottleneck`.

First, I set up a virtual environment. Then
```pip install bottleneck```
gives me this error:
```
Collecting bottleneck
  Using cached https://files.pythonhosted.org/packages/b8/d9/2b0beb43e0f8b25624699f918d6b00e53c46812da911eb15ec9ffebf1917/Bottleneck-1.3.0.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/alex/Projects/virtualenvs/pandas-dev/bin/python /home/alex/Projects/virtualenvs/pandas-dev/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpfzm0irx4
         cwd: /tmp/pip-install-hvrtun08/bottleneck
    Complete output (75 lines):
    running dist_info
    creating /tmp/pip-modern-metadata-jm_6d26f/Bottleneck.egg-info
    writing /tmp/pip-modern-metadata-jm_6d26f/Bottleneck.egg-info/PKG-INFO
    writing dependency_links to /tmp/pip-modern-metadata-jm_6d26f/Bottleneck.egg-info/dependency_links.txt
    writing requirements to /tmp/pip-modern-metadata-jm_6d26f/Bottleneck.egg-info/requires.txt
    writing top-level names to /tmp/pip-modern-metadata-jm_6d26f/Bottleneck.egg-info/top_level.txt
    writing manifest file '/tmp/pip-modern-metadata-jm_6d26f/Bottleneck.egg-info/SOURCES.txt'
    Traceback (most recent call last):
      File "/tmp/pip-build-env-40hp1wwq/overlay/lib/python3.8/site-packages/numpy/core/__init__.py", line 16, in <module>
        from . import multiarray
    ImportError: /tmp/pip-build-env-40hp1wwq/overlay/lib/python3.8/site-packages/numpy/core/multiarray.cpython-38-x86_64-linux-gnu.so: undefined symbol: cblas_cdotc_sub
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
[...]

PR #306, opened 2020-01-07
Use oldest-supported-numpy in pyproject.toml and explicitly require Python 3.6

astrofrog:
Currently the pyproject.toml file is missing a Numpy version for Python 3.8. Rather than add this, I've switched this file to use the [oldest-supported-numpy](https://github.com/scipy/oldest-supported-numpy) meta-package which is equivalent to specifying all the environment markers (see the README for that project) but means that the pyproject.toml file here doesn't need to be updated in future when a new version of Python is released since this just requires a new release of oldest-supported-numpy.

It looks like bottleneck requires Python 3.6 based on the classifiers, so I've also added an explicit ``python_requires`` which will prevent older Python versions from downloading newer releases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,14 +2,5 @@
 requires = [
     "setuptools",
     "wheel",
-    "numpy==1.13.3; python_version=='2.7' and platform_system!='AIX'",
-    "numpy==1.13.3; python_version=='3.5' and platform_system!='AIX'",
-    "numpy==1.13.3; python_version=='3.6' and platform_system!='AIX'",
-    "numpy==1.14.5; python_version=='3.7' and platform_system!='AIX'",
-    "numpy==1.17.3; python_version>='3.8' and platform_system!='AIX'",
-    "numpy==1.16.0; python_version=='2.7' and platform_system=='AIX'",
-    "numpy==1.16.0; python_version=='3.5' and platform_system=='AIX'",
-    "numpy==1.16.0; python_version=='3.6' and platform_system=='AIX'",
-    "numpy==1.16.0; python_version=='3.7' and platform_system=='AIX'",
-    "numpy==1.17.3; python_version>='3.8' and platform_system=='AIX'",
+    "oldest-supported-numpy"
 ]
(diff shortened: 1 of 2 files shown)


qwhelan:
Thanks for submitting this - I unfortunately won't have bandwidth to dig in until the weekend as I'm on call this week. One thing I'd like to make sure is that the test suite can reproduce the issue before merging this PR.

qwhelan:
Unfortunately I'm unable to reproduce the Arch error with the current approach in #307 so it's unclear if this resolves #277.

I'm inclined to merge as I'd like to not have to manually sync the pyproject.toml file with scipy/pandas - if you can give any additional insight into how to test that #277 is fixed with this approach, I'd appreciate it.

qwhelan:
@astrofrog Are there any other projects using this or planning to start doing so?

Let's say I release `bottleneck 1.4.0` with `py36` and `numpy 1.16` being in the supported window. Two years from now, they've fallen out of support. Can I build `bottleneck 1.4.0` with `py36` and `numpy 1.16` if this is a dependency?

astrofrog:
@qwhelan - yes this is used by e.g. astropy for example:

https://github.com/astropy/astropy/blob/master/pyproject.toml#L7

There are a number of other packages using this too:

https://github.com/search?l=TOML&q=oldest-supported-numpy&type=Code

Just to be clear, this package doesn't imply that this will be the oldest supported numpy right now, but the oldest version of numpy that works for each Python version. See also the note here:

https://github.com/scipy/oldest-supported-numpy/blob/master/setup.cfg#L17

As you can see, for some of the older Python versions, Numpy 1.13 is used (even though it's not supported right now). In other words, the version of Numpy corresponding to each Python version in the meta-package will never change. The only thing that will change is that new Python versions will be added, along with their corresponding first supported numpy version. Does this make sense?

astrofrog:
See also the FAQ here: https://github.com/scipy/oldest-supported-numpy

qwhelan:
Thanks @astrofrog and apologies for the delay!

astrojuanlu:
Looks like this PR was never released, opened #388 about it.

PR #369, opened 2021-01-06
use bool instead of np.bool

mathause:
numpy 1.20.0 deprecates `np.bool` etc, see.: https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated. I found one usage of this in bottleneck, that causes some [warnings in xarray](https://github.com/pydata/xarray/runs/1653359233#step:7:369).


qwhelan:
@mathause Thanks for fixing!

PR #378, opened 2021-03-22
Fix get_python_header_include() for Python 3.10

vstinner:
Don't use sys.version[:3] which doesn't work on "3.10.0a6 (...)"
string.


janezd:
Any chance of merging this and making a release?

Eager greetings from
&nbsp;&nbsp;&nbsp;&nbsp;https://github.com/biolab/orange3

(which depends on bottleneck and would like to support Python 3.10. :)).

markotoplak:
@qwhelan, are you still maintaining this package? If so, could you consider this (or similar) PR? Thanks.

vstinner:
Thanks ;-)

Issue #381, opened 2021-06-01
[BUG] Segfault or wrong value on array with non-vanilla strides.


lusewell:
**Describe the bug**
Segfault or wrong answer when running various functions on a numpy array with unusual strides. Looks like reading outside the array.

**To Reproduce**
```python
import bottleneck
import numpy as np
data = np.zeros((1,500, 2)).transpose(1,2,0)
bottleneck.nanmax(data) # This inconsistently either segfaults or gives a non-zero very large number.
```
bottleneck = 1.3.2
numpy = 1.16.6
python = 3.7.7

Operating system is RHEL7.

I've also replicated on ubuntu with python 3.9 and numpy 1.20, (still bottleneck 1.3.2) so doesn't seem to be too sensitive to versions of other packages.

**Expected behavior**
The nanmax of this data should always be zero.

**Additional context**
This in turn breaks `xarray.DataArray.max`

PR #382, opened 2021-06-02
Fix segfaulting issue with reduce iterator

lusewell:
Edit is much smaller than diff suggests - mostly just whitespace due to an extra layer of indentation.

The test added generates the segfault described in https://github.com/pydata/bottleneck/issues/381 without the rest of the change, and then the other change fixes this.

### DIFF
--- a/bottleneck/include/iterators.h
+++ b/bottleneck/include/iterators.h
@@ -98,37 +98,59 @@ static inline void init_iter_all(iter *it, PyArrayObject *a, int ravel, int anyo
         it->ndim_m2 = -1;
         it->length = 1;
         it->astride = 0;
-    } else if (C_CONTIGUOUS(a) && !F_CONTIGUOUS(a)) {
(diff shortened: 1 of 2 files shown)


lusewell:
This condition only actualyl needs to be that the last non-zero-strided axis is also the smallest non-zero-strided axis, but this was simpler to write.

rdbisme:
Hello I rebased this to try to merge it but the test segfaults in the CI...

lusewell:
Removing the change to iterators.h, the tests still cause a segfault - ie current state is broken.

The change of reverting the previous release means its quite a lot of effort to work out what the correct fix would now be.

rdbisme:
Hello, the reason is explained here: https://github.com/pydata/bottleneck/issues/388. We needed to release some fixes and the state of the main branch was broken.

The previous state is in stored in `develop`. Feel free to cherry pick the commit you need to fix this.

PR #403, opened 2022-02-22
BUG: install test data

mgorny:
Fix the setup script to install the data for tests.  This is necessary
for test_template.py to pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,14 +30,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install as develop
+      - name: Install
         run: |
-          pip install -e .
+          pip install .
 
       - name: Test with pytest
         run: |
           pip install pytest
-          pytest
+          pytest --pyargs bottleneck
 
   check:
     needs: test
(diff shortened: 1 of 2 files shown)


rdbisme:
Hello, could you add a test somewhere to trigger this? Because the tests are passing in the CI using `pytest` on the local repository. Or, are tests really needed to be shipped with the package?

mgorny:
> Hello, could you add a test somewhere to trigger this? Because the tests are passing in the CI using `pytest` on the local repository.

As in, a test that triggers the actual failure? Or should I try to modify the CI to test on of the installed package to trigger this?

> Or, are tests really needed to be shipped with the package?

I don't have an opinion either way. They were shipped originally, so I assumed that this is desirable.

rdbisme:
> As in, a test that triggers the actual failure? Or should I try to modify the CI to test on of the installed package to trigger this?
The one that is faster :)

mgorny:
Done. The first commit switches to non-develop install and uses `pytest --pyargs` to run tests from installed package rather than the current directory. It [breaks tests](https://github.com/mgorny/bottleneck/actions/runs/1884120856).

The second commit [fixes them afterwards](https://github.com/mgorny/bottleneck/actions/runs/1884136193).

rdbisme:
Thanks!

PR #427, opened 2023-01-24
CI: Add macos arm64 wheels to release build

ales-erjavec:
Add macos arm64 wheels to the release build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.7.0
         env:
           CIBW_SKIP: pp*
+          CIBW_ARCHS_MACOS: "x86_64 arm64"
 
       - name: Store wheel artifacts
         uses: actions/upload-artifact@v3


rdbisme:
Hello @ales-erjavec. Did you test the build? Because I thought the build doesn't work on `arm64`. Last time I checked, there was not an easy way of testing it on Github actions. IF there is, first step would be to enable testing on it.

rgommers:
I think this was automatically fixed because the wheel build job (https://github.com/pydata/bottleneck/blob/master/.github/workflows/ci.yml#L56) uses the defaults from `cibuildwheel` plus an `os:` matrix which contains `macos-latest`. When GitHub Actions moved to macOS 14 the runners changed to arm64, so now `bottleneck` has `macosx-arm64` wheels on PyPI but no longer any `x86-64` wheels. Maybe that's not too bad anymore, because arm64 Macs are 5 years old by now - most devs will be on arm64.

This PR can either be updated to add x86-64 on `macos-13` runners (already deprecated though, GitHub Actions will drop them at some point) or closed.

PR #436, opened 2023-08-19
Christoph Gohlke wheels have moved to github

stonebig:


rdbisme:
The pre-compiled wheel is now upload to PyPI for Windows too, so I guess this information is stale. Now just doing `pip install bottleneck` is enough to install bottleneck for most of platforms (still missing macOS support). 

Care to adapt this PR to reflect current situation instead?

stonebig:
ha ? Christoph only wrote "The wheels can be downloaded from the [Releases](https://github.com/cgohlke/numpy-mkl-wheels/releases) page."

Issue #439, opened 2024-01-06
[BUG]would it be possible to have Python-3.12 wheels ?


stonebig:
**Describe the bug**
Whish of Windows Python-3.12 wheel

RNarayan73:
Following error occurs on installing in Python 3.12 environment:
```
(skl312) PS C:\Users\naray\Python\venvs\skl312> pip install bottleneck

Collecting bottleneck
  Downloading Bottleneck-1.3.7.tar.gz (103 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 103.1/103.1 kB 3.0 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: numpy in c:\users\naray\python\venvs\skl312\lib\site-packages (from bottleneck) (1.26.3)
Building wheels for collected packages: bottleneck
  Building wheel for bottleneck (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for bottleneck (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [57 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build\lib.win-amd64-cpython-312
      creating build\lib.win-amd64-cpython-312\bottleneck
      copying bottleneck\_pytesttester.py -> build\lib.win-amd64-cpython-312\bottleneck
      copying bottleneck\_version.py -> build\lib.win-amd64-cpython-312\bottleneck
      copying bottleneck\__init__.py -> build\lib.win-amd64-cpython-312\bottleneck
      creating build\lib.win-amd64-cpython-312\bottleneck\benchmark
      copying bottleneck\benchmark\autotimeit.py -> build\lib.win-amd64-cpython-312\bottleneck\benchmark
      copying bottleneck\benchmark\bench.py -> build\lib.win-amd64-cpython-312\bottleneck\benchmark
      copying bottleneck\benchmark\bench_detailed.py -> build\lib.win-amd64-cpython-312\bottleneck\benchmark
      copying bottleneck\benchmark\__init__.py -> build\lib.win-amd64-cpython-312\bottleneck\benchmark
      creating build\lib.win-amd64-cpython-312\bottleneck\slow
      copying bottleneck\slow\move.py -> build\lib.win-amd64-cpython-312\bottleneck\slow
[...]

PR #440, opened 2024-01-06
Update ci.yml to get python-3.12

stonebig:
attempt to solve https://github.com/pydata/bottleneck/issues/439


rdbisme:
I think `macos-10.15` doesn't ship python12?

alimcmaster1:
I think we should close this in favour of: https://github.com/pydata/bottleneck/pull/442

PR #441, opened 2024-01-23
remove 'requires' metadata field deprecated for ~19 years

cburroughs:
The `Requires` field was deprecated with [PEP
345](https://peps.python.org/pep-0345/#summary-of-differences-from-pep-314) circa 2005.  [PEP
566](https://peps.python.org/pep-0566/#summary-of-differences-from-pep-345) circa 2017 introduced version 2.1 (used by bottleneck) of the metadata specification without any mention of `Requires`.

This is a problem because a tool working with Python packages might reasonably validate that the fields in the package metadata are valid per the specification.


cburroughs:
On the actions failures, I just see

image

so I'm not sure what to debug there.

PR #449, opened 2024-05-04
1.4.0rc4

rdbisme:


astrofrog:
It is building with Numpy 2.0rc1 but the test failures above are due to a change in API in Numpy 2.0 specifically the meaning of copy. But I can confirm that the wheels are built against Numpy 2.0 as far as I can see.

astrofrog:
In Numpy 2.0, you should use ``copy=None`` to mean copy if possible - ``copy=False`` will now error if it can't copy the array.

rdbisme:
Care to push a PR which fixes the API change (and maybe edits the tests correctly?)

PR #450, opened 2024-05-09
Test with numpy 2.0.0rc1 and fix API changes of np.array()'s copy kwarg

bnavigator:
Follow up of https://github.com/pydata/bottleneck/pull/449#issuecomment-2100314258

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        numpy-version: [""]
         architecture: [x86, x64]
         os:
           [
@@ -27,6 +28,12 @@ jobs:
             architecture: x86
           - os: macos-12
             architecture: x86
+        include:
+          - python-version: "3.12"
+            numpy-version: ">=2.0.0rc1"
+            architecture: x64
+            os: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v4
 
@@ -42,7 +49,7 @@ jobs:
 
       - name: Test with pytest
         run: |
-          pip install pytest
+          pip install pytest "numpy${{ matrix.numpy-version }}"
           pytest --pyargs bottleneck
 
   check:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -42,15 +49,9 @@ jobs:
...
-          pip install pytest
+          pip install pytest "numpy${{ matrix.numpy-version }}"
           pytest --pyargs bottleneck
 
-  check:
-    needs: test
-    runs-on: ubuntu-latest
-    steps:
-      - name: Placeholder for CI checks in PRs
-        run: echo "Done"
```
rdbisme, line 53: I think this is needed to avoid dangling checks in the PRs.

.github/workflows/ci.yml
```diff
@@ -42,15 +49,9 @@ jobs:
 
       - name: Test with pytest
         run: |
-          pip install pytest
+          pip install pytest "numpy${{ matrix.numpy-version }}"
```
rdbisme, line 52: Why not in the install step?

.github/workflows/ci.yml
bnavigator: How? It's just another superflous job next to the real `test` jobs.

.github/workflows/ci.yml
bnavigator: Because it ist part of the test setup. The install/build step is separate. That one actually tests the installability from the runtime requirements,

https://github.com/pydata/bottleneck/blob/787d6daa292ef013efb2ce93f100079457330363/setup.py#L195

which installs numpy 1.26 at the moment.

https://github.com/pydata/bottleneck/actions/runs/9022389505/job/24849379279?pr=450#step:4:31

mgorny:
Thanks for this!

.github/workflows/ci.yml
rdbisme: I think that's needed. Otherwise when we remove other tests (e.g. deprecating python versions), they will be dangling in the "missing" checks of the PR (as you can see right now for the `check` itself). 

I added the `check` to avoid changing the required checks everytime. I made the PR depends on it, such that whatever upstream step of the pipeline change, the PR will keep dependending on check and works as expected.

bnavigator: I still think that there would be better ways to achieve what you are trying to accomplish with that dummy job, but whatever. This PR is not the place to discuss this. I reverted the removal.

rdbisme: Happy to receive a PR to fix that. I'm not a Github expert, so I might surely missing something :)

rdbisme:
Thanks a lot.

Issue #394, opened 2021-12-02
Build failing for arm64 


odidev:
I have been working on installing and testing bottleneck package for amd64 and arm64 architectures. I am getting the below error on arm64 architecture for the command **python setup.py build_ext --inplace** 
```
aarch64-linux-gnu-gcc -pthread -E -Wdate-time -D_FORTIFY_SOURCE=2 -I/usr/include/python3.8 -I/usr/include/python3.8 -o _configtest.i _configtest.c 
_configtest.c:1:10: fatal error: emmintrin.h: No such file or directory 

    1 | #include <emmintrin.h> 
      |          ^~~~~~~~~~~~~ 
compilation terminated. 
```
Further I explored about the issue, got that ARM CPUs do not support this.  

Could you please provide your feedback regarding this. 

**Log for reference:** https://github.com/odidev/bottleneck/runs/4380495013?check_suite_focus=true

odidev:
Is there any update for the above issue?

odidev:
@rdbisme Thanks for update.

blthayer:
Hello,

I am also interested in an ARM build for `bottleneck`. It looks like @odidev was able to get this to succeed:
- https://github.com/pydata/bottleneck/compare/master...odidev:master
- https://github.com/odidev/bottleneck/runs/5536020987?check_suite_focus=true

Is there any plan for official `bottleneck` support for ARM?

rdbisme:
Hello @blthayer, I think it's for sure interesting to have support for ARM as soon as it doesn't impact  performance on other architectures. Pull requests welcome! :) I'd say to split the contribution as: 
1. Adding missing code needed to allow build on `armv7` and `aarch64` and Apple Silicon
2. Ensure that the added code doesn't impact other architectures
3 Tweaking CI to test, build and releasse for ARM

PR #464, opened 2024-10-24
Publish linux arm64 wheels

mark-thm:
We'd like to use bottleneck on Linux arm64 machines and it'd be great to consume a wheel instead of an sdist for that.

Following the guide at https://cibuildwheel.pypa.io/en/stable/faq/#emulation, cibuildwheel supports QEMU emulation to make the arm64 build work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
-        architecture: [x86, x64, arm64]
+        architecture: [x86, x64]
         os:
           [
             ubuntu-latest,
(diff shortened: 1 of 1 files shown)


rdbisme:
Thanks. 

I'm reasonably convinced this still needs to be fixed: https://github.com/pydata/bottleneck/issues/394

mark-thm:
Oh, interesting -- I'm able to successfully install (and thus build) from sdist on Ubuntu 22.04 running on an arm64 machine.

rdbisme:
Ok! Any chance you might contribute also in the tests section of the CI? That will allow us to be sure it works. I'm not 100% sure on how to do that, I hope Github actions has Arm64 runners.

mark-thm:
ah, sorry, I see that abbreviation is different than I expected

mark-thm:
Running some experiments on another repo about getting Python tests running on arm64 Linux runners but not having much success. GitHub is not yet supporting Linux/arm64 runners for OSS projects, so I’m not able to do this natively.

I’m confident in the cibuildwheel support (just followed their docs).

Some ideas here:
- run the cibuildwheel build minus the publishing component in CI (don’t think this will run your tests but it will show that the library successfully builds)
- run the tests in a qemu emulated docker container
- don’t modify the test portion

groodt:
GitHub hopefully will have Arm64 for OSS by end of year. https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/

So hopefully that holds true.

CCardosoDev:
Also interested on this issue, any updates? 😊

rgommers:
Linux arm64 runners on GitHub Actions have been pretty stable, we've been using them in NumPy and SciPy for a while now. It'd be nice to update this PR to use them and remove QEMU usage. That would probably make this a lot easier to get merged.

PR #468, opened 2025-02-18
ENH: enable CPython free-threading support

andfoy:
This PR enables free-threading builds of bottleneck, available from Python 3.13 onwards, this new distribution allows Python packages to leverage true parallel support, as the GIL is now optional. 

In particular, this PR performs the following checks:
- [x] Test the package using `pytest-run-parallel`, which allows pytest suites to be run concurrently (per test). This analysis didn't throw any major concurrency-related issues, the only tests that were marked as thread-unsafe were related to memory usage checking and concurrent file overwriting.
- [x] Add a CI to run tests in parallel, as well to build and release free-threaded wheels
- [x] Check package compliance using Thread Sanitizer (TSAN)
- [x] Marking each C module as free-threaded compatible

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
 
 
       - if: ${{ endsWith(matrix.python-version, 't') }}
-        run:
+        run: |
           pip install pytest-run-parallel
           echo "PYTEST_ADDOPTS=--parallel-threads=4" >> "$GITHUB_ENV"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.23.2
         env:
-          CIBW_SKIP: pp*
+          CIBW_SKIP: "pp* *t-manylinux_i686 *t-musllinux_i686"
           CIBW_ENABLE: cpython-freethreading
 
       - name: Store wheel artifacts
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.23.2
         env:
           CIBW_SKIP: pp*
+          CIBW_ENABLE: cpython-freethreading
 
       - name: Store wheel artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"]
         architecture: [x86, x64]
         os:
           [
@@ -29,10 +29,25 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python-version }}
+        if: ${{ matrix.python-version != '3.13t' }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           architecture: ${{ matrix.architecture }}
+
+      - name: Set up free-threaded Python ${{ matrix.python-version }}
+        if: ${{ matrix.python-version == '3.13t' }}
+        uses: astral-sh/setup-uv@b5f58b2abc5763ade55e4e9d0fe52cd1ff7979ca
+        with:
+          python-version: ${{ matrix.python-version }}
+          enable-cache: false
+
+      - if: ${{ matrix.python-version == '3.13t' }}
+        run:
+          uv pip install --python=${{ matrix.python-version }} pip
+          uv pip install --python=${{ matrix.python-version }} --pre -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
+          uv pip install --python=${{ matrix.python-version }} pytest-run-parallel
+          echo "PARALLEL_THREADS=4" >> "$GITHUB_ENV"
 
       - name: Install
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,24 +29,16 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python-version }}
-        if: ${{ matrix.python-version != '3.13t' }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@9e62be81b28222addecf85e47571213eb7680449
         with:
           python-version: ${{ matrix.python-version }}
           architecture: ${{ matrix.architecture }}
 
-      - name: Set up free-threaded Python ${{ matrix.python-version }}
-        if: ${{ matrix.python-version == '3.13t' }}
-        uses: astral-sh/setup-uv@b5f58b2abc5763ade55e4e9d0fe52cd1ff7979ca
-        with:
-          python-version: ${{ matrix.python-version }}
-          enable-cache: false
 
-      - if: ${{ matrix.python-version == '3.13t' }}
+      - if: ${{ endsWith(matrix.python-version, 't') }}
         run:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -32,19 +32,26 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@9e62be81b28222addecf85e47571213eb7680449
```
rgommers, line 35: This should no longer be needed, since the latest `setup-python` release now includes support. Revert this change?

.github/workflows/ci.yml
```diff
@@ -32,19 +32,26 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
           architecture: ${{ matrix.architecture }}
 
+
+      - if: ${{ endsWith(matrix.python-version, 't') }}
+        run:
+          pip install --pre -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
```
rgommers, line 43: This can be left out, since `numpy` 2.2.4 has stable support for free-threading. You can just drop this line I believe.

.github/workflows/ci.yml
```diff
@@ -32,19 +32,26 @@ jobs:
...
+
       - name: Install
         run: |
           pip install .
 
       - name: Test with pytest
         run: |
           pip install pytest
-          pytest --pyargs bottleneck
+          pytest --parallel-threads=$PARALLEL_THREADS --pyargs bottleneck
```
rgommers, line 54: This doesn't look correct, since `pytest-run-parallel` is only installed for cp313t. On other Python versions, this should result in:
```
pytest: error: unrecognized arguments: --parallel-threads=
```

It's annoying that CI doesn't run on this PR, but can you try running the CI job on your own fork to make sure it passes? (same for wheel builds)

rgommers (commented):
Thanks for working on this @andfoy. Mostly looks good, a few comments.

rdbisme:
I have approved the workflows, so now the CI will run here!

rgommers (approved):
Thanks @andfoy, this LGTM now. The changes are very minimal, and I verified locally too that everything works as advertised, both with and without `pytest-run-parallel` installed.

andfoy:
I confirm wheels are building locally: https://github.com/andfoy/bottleneck/actions/runs/14230390834

andfoy:
I need to push a final commit, and it should be ready to go

rdbisme:
Why are we skipping `muslinux` x86 builds?

rgommers:
> Why are we skipping `muslinux` x86 builds?

I bet it's because `numpy` doesn't provide `cp313t_*_i686` wheels (see https://pypi.org/project/numpy/2.2.4/#files), so (a) they won't be useful, and (b) building `numpy` for `cp3113t` in CI here can't be done fully automatically within the `cibuildwheel` invocation until there's a Cython 3.1.0 release.

andfoy:
> I bet it's because numpy doesn't provide cp313t_*_i686 wheels

This is the actual issue at hand, since the job is only failing on free-threaded builds under x86, I decided to skip them for now.

rgommers:
@andfoy the failure caught a problem. It looks like a cross-merge conflict because `filterwarnings = ["error"]` was added in gh-473. I think the underlying reason is that the `warnings.catch_warnings` usage is wrapped inside the `unit_maker` utility in `bottleneck/tests/reduce_test.py`, and hence `pytest-run-parallel` doesn't automatically run that test single-threaded. Could you verify, and also check that the same issue isn't present in other test files?

Keeping the traceback for posterity:

<details>

```
_____________________________ test_reduce[nanmean] _____________________________

func = <built-in function nanmean>

    @pytest.mark.parametrize("func", bn.get_functions("reduce"), ids=lambda x: x.__name__)
    def test_reduce(func):
        """test reduce functions"""
>       return unit_maker(func)

func       = <built-in function nanmean>

/opt/hostedtoolcache/Python/3.13.2/x64-freethreaded/lib/python3.13t/site-packages/bottleneck/tests/reduce_test.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

func = <built-in function nanmean>, decimal = 5, skip_dtype = ('nansum', 'ss')

    def unit_maker(func, decimal=5, skip_dtype=("nansum", "ss")):
        """Test that bn.xxx gives the same output as bn.slow.xxx."""
        fmt = "\nfunc %s | input %s (%s) | shape %s | axis %s | order %s\n"
        fmt += "\nInput array:\n%s\n"
        name = func.__name__
        func0 = eval("bn.slow.%s" % name)
        for i, a in enumerate(arrays(name)):
            if a.ndim == 0:
                axes = [None]  # numpy can't handle e.g. np.nanmean(9, axis=-1)
            else:
                axes = list(range(-1, a.ndim)) + [None]
            for axis in axes:
                actual = "Crashed"
                desired = "Crashed"
                actualraised = False
                try:
                    # do not use a.copy() here because it will C order the array
                    actual = func(a, axis=axis)
                except:  # noqa
                    actualraised = True
                desiredraised = False
                try:
                    with warnings.catch_warnings():
                        warnings.simplefilter("ignore")
                        desired = func0(a, axis=axis)
                except:  # noqa
                    desiredraised = True
                if actualraised and desiredraised:
                    pass
                else:
                    tup = (
                        name,
                        "a" + str(i),
                        str(a.dtype),
                        str(a.shape),
                        str(axis),
                        array_order(a),
                        a,
                    )
                    err_msg = fmt % tup
                    if actualraised != desiredraised:
                        if actualraised:
                            fmt2 = "\nbn.%s raised\nbn.slow.%s ran\n\n%s"
                        else:
                            fmt2 = "\nbn.%s ran\nbn.slow.%s raised\n\n%s"
                        msg = fmt2 % (name, name, traceback.format_exc())
                        err_msg += msg
>                       assert False, err_msg
E                       AssertionError: 
E                         func nanmean | input a80 (int32) | shape (0, 0) | axis 0 | order C,F
E                         
E                         Input array:
E                         []
E                         
E                         bn.nanmean ran
E                         bn.slow.nanmean raised
E                         
E                         NoneType: None
E                         
E                       assert False

a          = array([], shape=(0, 0), dtype=int32)
actual     = array([], dtype=float64)
actualraised = False
axes       = [-1, 0, 1, None]
axis       = 0
da         = dtype('float64')
dd         = dtype('float64')
decimal    = 5
desired    = 'Crashed'
desiredraised = True
err_msg    = '\nfunc nanmean | input a80 (int32) | shape (0, 0) | axis 0 | order C,F\n\nInput array:\n[]\n\nbn.nanmean ran\nbn.slow.nanmean raised\n\nNoneType: None\n'
fmt        = '\nfunc %s | input %s (%s) | shape %s | axis %s | order %s\n\nInput array:\n%s\n'
fmt2       = '\nbn.%s ran\nbn.slow.%s raised\n\n%s'
func       = <built-in function nanmean>
func0      = <function nanmean at 0x2eec1e3c3f0>
i          = 80
msg        = '\nbn.nanmean ran\nbn.slow.nanmean raised\n\nNoneType: None\n'
name       = 'nanmean'
skip_dtype = ('nansum', 'ss')
tup        = ('nanmean', 'a80', 'int32', '(0, 0)', '0', 'C,F', ...)

/opt/hostedtoolcache/Python/3.13.2/x64-freethreaded/lib/python3.13t/site-packages/bottleneck/tests/reduce_test.py:67: AssertionError
```

</details>

andfoy:
I confirm that `bottleneck/tests/nonreduce_axis_test.py` also recycles the `unit_maker` as well, however, they can be marked automatically as thread-unsafe when using https://github.com/Quansight-Labs/pytest-run-parallel/pull/37

rgommers:
@andfoy now that the new `pytest-run-parallel` release is out, can you update this?

rgommers:
This is all happy now. I ran the wheel builds again on my fork for completeness, all green too: [CI log](https://github.com/rgommers/bottleneck/actions/runs/14735525996/job/41360522521). Time to get this merged. Thanks @andfoy!

Issue #467, opened 2024-12-07
Readme is not up to date regarding Python support


ogencoglu:
It seems to support Python3.12 but Readme does not mention it. It is also unclear if it supports numpy 2.x.x.

rdbisme:
Hello, thanks @ogencoglu. Feel free to provide a PR! :)

PR #469, opened 2025-03-26
Update Python version support and install docs in README

rgommers:
Closes gh-467

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,10 +107,18 @@ and LICENSES directory for details.
 Install
 =======
 
+Bottleneck provides binary wheels on PyPI for all the most common platforms.
+Binary packages are also available in conda-forge. We recommend installing binaries
+with ``pip``, ``uv``, ``conda`` or similar - it's faster and easier than building
+from source.
+
+Installing from source
+----------------------
+
 Requirements:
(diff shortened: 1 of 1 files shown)


rdbisme:
Nice. There's a related issue to this: I never managed to get the hands on the RTD account. It still works, but I don't control it. Any experience on what we might do to get back access to it?

rgommers:
> Nice. There's a related issue to this: I never managed to get the hands on the RTD account. It still works, but I don't control it. Any experience on what we might do to get back access to it?

I don't have experience with that, aside from migrating away from RTD to GitHub Pages. https://docs.readthedocs.com/platform/stable/abandoned-projects.html#reachability doesn't look like there's a process for this, aside from getting in touch with the current owner of the RTD login credentials for `bottleneck`. I assume you already tried that and got no response? If so, it may be worth asking RTD [site support](https://docs.readthedocs.com/platform/stable/support.html), since it's a little odd that this question isn't answered in their docs (unless I missed it of course).

rgommers:
Shall we merge this and open a new issue for the RTD permissions? It feels like that will take time to work through, and isn't directly related.

PR #472, opened 2025-03-28
Fix broken CI, source dir was picked up instead of installed package

rgommers:
In addition:

- Make the `pip install .` command verbose, so any build warnings show up in CI log - helps with debugging when anything fails.
- Remove dummy job with no content
- Bump version of cibuildwheel to (latest) 2.23.2
- Remove `macos-12` run, it's deprecated and no longer starts or with a delay, and it was redundant with macos-13.


rgommers:
Ah wait, the dummy `check` job is there as a gate for merging, so it actually has a function:

image

I'll put it back and tweak the comment on why it's there.

rgommers:
This fixes CI and is enough to make other PRs mergeable again (the pre-commit failure isn't blocking).

@rdbisme thank you for the commit bit! If you have any preferences on how to use my newly acquired powers, please let me know - in particular, do you want to review all PRs from me, or are you fine with me self-merging PRs that are business as usual, and only wait for your review on things that are actually consequential (e.g., dropping a Python version, adding a new function, etc.)?

.github/workflows/ci.yml
```diff
@@ -39,14 +36,18 @@ jobs:
...
       - name: Test with pytest
         run: |
           pip install pytest
+          cd doc  # avoid picking up bottleneck from the source dir
           pytest --pyargs bottleneck
 
   check:
+    # This job is here is the "Required" one for merging PRs, and
+    # it only runs after all the `test` jobs above have run. Hence
+    # it serves as a check that CI actually ran before a PR gets merged.
```
rdbisme, line 50: There is also another usecase: when you remove jobs (e.g. obsolete python), they stay there "dangling" and you need to remove them manually. If you make the MR depend on this job instead, then you don't need to do that anymore.

rdbisme:
> This fixes CI and is enough to make other PRs mergeable again (the pre-commit failure isn't blocking).
> 
> @rdbisme thank you for the commit bit! If you have any preferences on how to use my newly acquired powers, please let me know - in particular, do you want to review all PRs from me, or are you fine with me self-merging PRs that are business as usual, and only wait for your review on things that are actually consequential (e.g., dropping a Python version, adding a new function, etc.)?

Sure. Let's do a couple round trips and then you'll be "free" to go on your own :). 

Thanks for taking the time and giving me a hand with this project.

rdbisme:
Also this project historically was holding a sort of (unenforced) semantic commit style for commits. Do we want to keep it? In that case we might introduce pre-commits again.

rgommers:
Sounds good!

> Also this project historically was holding a sort of (unenforced) semantic commit style for commits. Do we want to keep it? In that case we might introduce pre-commits again.

I personally avoid pre-commit, too many security risks. If a project I contribute to uses it, I just look at what fails in CI and fix it up. 

Happy to use the `MNT:` etc. prefixes of course - I just missed that that was being used in this repo.

PR #473, opened 2025-04-01
MAINT: update maximum numpy version to build against

rgommers:
The rule is <N+3, and numpy 2.2.x is out, so the bound should be <2.5.

Also add a `pytest.ini` file that will turn deprecation warnings in the test suite into errors. This will catch any deprecations from `numpy` or `python`, so that they can be fixed in time. Having the test suite run deprecation-warning-free is the guarantee that the <N+3 upper bound will work. The `pytest.ini` contents are derived from what NumPy and SciPy also use.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,12 @@ requires = [
     #   2. Note that building against numpy 1.x works fine too - users and
     #      redistributors can do this by installing the numpy version they like
     #      and disabling build isolation.
-    #   3. The <2.3 upper bound is for matching the numpy deprecation policy,
-    #      it should not be loosened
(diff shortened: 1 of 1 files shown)


rdbisme:
Can we maybe just use  `pyproject.toml`. No big opinions here, just maybe having the same syntax?

rgommers:
Ah for `pytest` config (https://doc.pytest.org/en/latest/reference/customize.html#pyproject-toml). Yes, we can - that does seem nicer indeed. That support is relatively new, so I didn't think of it. Will update.

We should also be using `pyproject.toml`'s `[project]` section for project metadata instead of hiding it in `setup.py`, but I'll leave that for another time.

PR #475, opened 2025-04-29
CI: bump ubuntu-20.04 to 22.04 in CI jobs

rgommers:
20.04 is now failing, it is past the end of its announced deprecation period, see https://github.com/actions/runner-images/issues/11101


rgommers:
For completeness, the failures on gh-304:

image

This is a trivial fix, so I'll hit the green button straight away.

PR #395, opened 2021-12-31
Mac M1 Support

itsjohnward:
Incorporated @NikZak's suggestion from https://github.com/pydata/bottleneck/issues/385#issuecomment-928035398 and also removed the emmintrin header on arm64 ARM per https://reviews.llvm.org/D109686#change-qMIOtz4d3wgp.

Tested like so:

```
$ SDKROOT=$(xcrun --sdk macosx --show-sdk-path) pip install -e .[test] && pytest .
...
===================================== test session starts ======================================
platform darwin -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/john/bottleneck, configfile: setup.cfg
plugins: hypothesis-6.34.1
collected 209 items                                                                            

bottleneck/tests/input_modification_test.py .............................                [ 13%]
bottleneck/tests/list_input_test.py .............................                        [ 27%]
bottleneck/tests/memory_test.py .                                                        [ 28%]
bottleneck/tests/move_test.py .................................                          [ 44%]
bottleneck/tests/nonreduce_axis_test.py .........................                        [ 55%]
bottleneck/tests/nonreduce_test.py ..............                                        [ 62%]
bottleneck/tests/reduce_test.py ........................................................ [ 89%]
...                                                                                      [ 90%]
bottleneck/tests/scalar_input_test.py ..................                                 [ 99%]
bottleneck/tests/test_template.py .                                                      [100%]

======================================= warnings summary =======================================
bottleneck/tests/scalar_input_test.py::test_scalar_input[ss]
  /Users/john/bottleneck/bottleneck/slow/reduce.py:87: RuntimeWarning: overflow encountered in multiply
    y = np.multiply(a, a).sum(axis)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=============================== 209 passed, 1 warning in 32.31s ================================
```

Issue #474, opened 2025-04-03
Next release, with more supported platforms


rgommers:
I'd like to work through the platform support related open PRs and issues, get them merged/addressed, and then tag a new release. 

Platforms to be added compared to the current 1.4.2 release for which there are wheels on PyPI:
- Linux `aarch64` (new)
- macOS x86-64 (this got lost but used to be there)
- free-threaded CPython (new)

In addition, it looks like it's time to drop Python 3.9, either before or after this release. NumPy has dropped 3.9 a long time ago, and has even already droppped 3.10 in `main`. [SPEC 0](https://scientific-python.org/specs/spec-0000/) says that we're already halfway through the 3.11 support window right now. Dropping 3.9 will also be helpful to stop the wheel build matrix from growing a lot. Around September/October we'll need to support Python 3.14, and we may want to consider doing 3.11-3.14 at that point.

@rdbisme does that sound good to you? And would you prefer for that release to be 1.5.0 or 1.4.3? And drop 3.9 now, or after the next release?

rgommers:
> [@rdbisme](https://github.com/rdbisme) does that sound good to you? And would you prefer for that release to be 1.5.0 or 1.4.3? And drop 3.9 now, or after the next release?

Any opinion on this one? I think things are ready for a next release now that gh-476 is merged.

rdbisme:
Hello, sorry for missing this. Considering the free-threading support, I'd say 1.5.0? 
For 3.9 we can remove support after this release.

rgommers:
That sounds good to me. Let me open a PR with an update to `RELEASE.rst`. It seems like the rest of the release process is pushing a tag and uploading wheels and sdist to PyPI (after downloading from the last CI run locally I think?). Can you take care of that? I'm happy to help if needed, but would need PyPI permissions for it.

rdbisme:
This is done: https://pypi.org/project/Bottleneck/1.5.0/

rgommers:
Great, thanks!

PR #476, opened 2025-05-09
CI: add macOS arm64, Linux aarch, remove i686 wheels, trim test matrix

rgommers:
This modernizes the CI setup to do the following:

- Test jobs that runs on PRs:
    - Adds macOS arm64
    - Adds Linux aarch64
    - Removes running duplicate jobs on `ubuntu-22.04` and `ubuntu-latest` (both x86-64, not useful)
    - Removes running jobs for every single minor Python version, but rather runs mininum (3.9), maximum (3.13) supported versions + free-threading (3.13t)
    - Refactors the test matrix so it's easier to read (no more  `exclude`) jobs
- Wheel build jobs (logs from a wheel build run [here](https://github.com/rgommers/bottleneck/actions/runs/14933404774/job/41955095607)):
    - Restores macOS x86-64 (those went missing from recent releases due to `macos-latest` switching from x86-64 to arm64, see e.g. https://pypi.org/project/Bottleneck/1.4.2/#files)
    - Adds Linux aarch64
    - Removes 32-bit Linux jobs (those are super slow and not useful, because NumPy stopped shipping i686 wheels a while back

I left the `windows-2019`/`windows-latest` test jobs duplication; that's probably also not useful, but it's harder to be sure of that than on Linux. Windows arm64 can be added later, once NumPy 2.3.0 publishes `win-arm64` wheels.

Adds the platforms as proposed in gh-474
Closes gh-464
Closes gh-395
Closes gh-394

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,28 +3,40 @@ name: Github Actions
 on: ["push", "pull_request"]
 
 jobs:
-  test:
+  test-linux-macos:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"]
-        architecture: [x86, x64]
+        python-version: ["3.9", "3.13", "3.13t"]
         os:
-          [
-            ubuntu-latest,
-            ubuntu-22.04,
-            macos-13,
-            windows-latest,
-            windows-2019,
-          ]
-        exclude:
-          - os: ubuntu-latest
-            architecture: x86
-          - os: ubuntu-22.04
-            architecture: x86
-          - os: macos-13
-            architecture: x86
+          # Note that macos-13 is x86-64 (deprecated already),
+          # and macos-latest is arm64.
+          [ubuntu-22.04, ubuntu-24.04-arm, macos-13, macos-latest]
+    steps:
+      - uses: actions/checkout@v4
 
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install
+        run: |
+          pip install . -v
+
+      - name: Test with pytest
+        run: |
+          pip install pytest
+          cd doc  # avoid picking up bottleneck from the source dir
+          pytest --pyargs bottleneck
+
+  test-windows:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.13", "3.13t"]
+        architecture: [x86, x64]
+        os: [windows-latest, windows-2019]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,28 +3,40 @@ name: Github Actions
...
+          cd doc  # avoid picking up bottleneck from the source dir
+          pytest --pyargs bottleneck
+
+  test-windows:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.13", "3.13t"]
+        architecture: [x86, x64]
+        os: [windows-latest, windows-2019]
```
rdbisme, line 39: I'm not really a huge fan of duplicating the steps for windows. I know it's a bit less readable using exclude, but I'd like to have a monolithic test suite to avoid divergence between the two test jobs.

.github/workflows/ci.yml
```diff
@@ -3,28 +3,40 @@ name: Github Actions
...
+        run: |
+          pip install pytest
+          cd doc  # avoid picking up bottleneck from the source dir
+          pytest --pyargs bottleneck
+
+  test-windows:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.13", "3.13t"]
```
rdbisme, line 37: Are you sure this is enough? I was of the opinion that, as far it doesn't have an impact because they run in parallel, we might want to test all the supported python versions

rgommers: It should be. Other option: how about a separate job on only `ubuntu-latest` then that exhaustively does all Python versions? That'd be 6 jobs.

It doesn't seem reasonable to do 6 platforms x 6 python versions x 2 architectures = 72 jobs.

.github/workflows/ci.yml
rgommers: That really doesn't work anymore. Some platforms don't have arm64, none but windows have x86. The list of excludes would be longer, and much harder to audit, then the duplicated job steps.

I'd say this is pretty clean. But if you don't like this, how about taking these steps and putting them in `.github/bottleneck_action/action.yml`? That would then contain:
```
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.python-version }}

      - name: Install
        run: |
          pip install . -v

      - name: Test with pytest
        run: |
          pip install pytest
          cd doc  # avoid picking up bottleneck from the source dir
          pytest --pyargs bottleneck
```

Then the jobs will look like:
```yaml
    steps:
      - uses: actions/checkout@v4
      - uses: ./.github/bottleneck_action
```
So no duplication. We do that in some jobs in NumPy, e.g. https://github.com/numpy/numpy/blob/e9e981e9d45335fd6b758e620812772e19143f35/.github/workflows/linux_simd.yml#L69

.github/workflows/ci.yml
rdbisme: Ok sounds good.

.github/workflows/ci.yml
rdbisme: Yes, that's cool. Thanks!

rgommers:
Comments addressed now - please take another look.

PR #481, opened 2025-08-30
MAINT: add Python 3.14, drop 3.9, fix some CI and build stuff

rgommers:
Python 3.14 is in the release candidate stage, meaning this is the time to add 3.14 wheels on PyPI.

Changes:
- Add Python 3.14
- Drop Python 3.9, it's end of life (I considered dropping 3.10 as well, because NumPy already did - but decided against that since it costs little to keep it)
- Fix CI for GHA dropping `windows-2019`
- Add free-threading trove classifier
- Bump `numpy` upper bound up

Also, two questions @rdbisme:
- Any preference for making the next release number 1.6.0 vs 1.5.1? Either way should work; I'd lean towards 1.6.0 because of dropping Python 3.9
- Support for Intel Macs will disappear on GHA starting 1 Sep: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down. Are you okay with just dropping those wheels, probably after doing one more release with it now? Seems like the easiest, and I expect that that is what many projects will do.


rdbisme:
> Any preference for making the next release number 1.6.0 vs 1.5.1? Either way should work; I'd lean towards 1.6.0 because of dropping Python 3.9
Yes, 1.6.0 sounds good to me

> Support for Intel Macs will disappear on GHA starting 1 Sep: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down. Are you okay with just dropping those wheels, probably after doing one more release with it now? Seems like the easiest, and I expect that that is what many projects will do.

I definitely trust your judgment here, you have way more experience with this. I was a user of macports, which tries to keep things going for old macs, but not having infrastructure to compile it on makes it difficult. I guess it's impossible to crosscompile on Mac Arm?

rdbisme:
> cibuildwheel: Invalid skip selector: 'pp*'. This selector matches a group that wasn't enabled. Enable it using the `enable` option or remove this selector. This selector will have no effect.

Apparently that skip selector is not needed anymore with new `cibuildwheel`?

PR #482, opened 2025-09-04
MNT: Prepare changelog for 1.6.0 

rdbisme:


rgommers:
@rdbisme I see you tagged 1.6.0rc1, it had a hiccup in CI yesterday due to GHA runner availability; it's up on PyPI now. It's safe to tag 1.6.0 soon as well I'd think, there is nothing in this release that should be fragile.

PR #483, opened 2025-09-05
MAINT: fix warning about `pp_*` skip selector for PyPy wheels not being needed

rgommers:
This is a follow-up to gh-481, which moved to cibuildwheel 3.x which removed PyPy from the default set of wheels being built.


rgommers:
This is a trivial fix, so I'll go ahead and merge it.

Issue #437, opened 2023-09-06
use move_mean and set window&min_count =1, diff result


gucasbrg:
```
import bottleneck as bn
[...]

PR #488, opened 2025-11-21
Fix move_mean to handle window=1 without floating-point errors

gliptak:
#437


gliptak:
@rgommers please review

PR #489, opened 2025-11-27
WHL: re-add macos-intel to build matrix

neutrinoceros:
Searching through the history I couldn't find why this target was dropped between 1.3.8 and 1.4.0rc4, so my current assumption is that it happened by accident, when the `macos-latest` image tag moved from an x86_64 runner to arm64. While I was surprise to find bottleneck needed to be built to install it on an old mac, it did build succesfully so I don't know of any blocking issue, so I'd like to propose re-introducing it on purpose.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest, ubuntu-24.04-arm]
+        os: 
+          - macos-15-intel
+          - macos-latest
+          - windows-latest
+          - ubuntu-latest
+          - ubuntu-24.04-arm
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -52,7 +52,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest, ubuntu-24.04-arm]
+        os: 
+          - macos-15-intel
+          - macos-latest
+          - windows-latest
+          - ubuntu-latest
+          - ubuntu-24.04-arm
```
neutrinoceros: while I was at it I also expanded this one liner to improve how useful `git blame` is in the future.

neutrinoceros:
It looks like wheels are not built as part of CI, and even then, I'm not sure if they're systematically tested. Happy to contribute in that area too if desired.

PR #491, opened 2025-12-10
MNT: move static project metadata to `pyproject.toml`

neutrinoceros:


neutrinoceros:
resolved by setting
```toml
[tool.setuptools]
include-package-data = false
```

rgommers (approved):
This looks like a significant improvement, CI is happy, and `diffoscope` doesn't show any nontrivial difference for sdist and wheel. So in it goes, thanks @neutrinoceros!

PR #493, opened 2025-12-10
MNT: remove `setup.cfg`

neutrinoceros:
- **MNT: move versioneer configuration to `pyproject.toml`**
- **CLN: drop unused pytest configuration from `setup.cfg`**
- **MNT: rename `setup.cfg` -> `.flake8`**

PR #497, opened 2026-01-14
CI: bump setup-python to v6 to resolve a missing win-arm64 zipfile

rgommers:


rgommers:
Well that didn't help, still the same windows-arm issue as in https://github.com/pydata/bottleneck/pull/493#issuecomment-3751357606.

neutrinoceros:
Also tried #500 , didn't help either.
I'd report this upstream if I even understood which part is likely to be broken.

PR #504, opened 2026-01-20
MNT/STY: migrate linting+formatting from flake8+black to ruff, fix existing violations

neutrinoceros:
Manual fixes are isolated to individual, ruleset-based commit. There's a single commit for all (safe) automated fixes, and I set `git blame` to ignore it in the future.
I also included 2 other automated commits containing fixes marked as "unsafe" (meaning ruff doesn't guarantee strict backward compat in the general case). I reviewed them manually and think they are correct and not breaking, but I still keep them separate so git blame correctly points to me as the latest editor or these lines, just in case it actually breaks anything.


neutrinoceros:
done

PR #512, opened 2026-02-01
DEP: commit `uv.lock` to repository and use locked env in CI

### LINKED ISSUE #508: CI: locking CI env
I would like to propose committing (a) lock file(s) to the repository and use them in CI, and have dependabot periodically update it (them).
This would reduce the chance that a change in the rest of the world breaks CI while we're not looking, and that each time this repo (infrequently) receives attention, we find something broken without any guarantee that it'll be easy to understand.
In contrast, if upgrades only happen deliberately (via dependabot or otherwise) and are isolated to specific PRs, it's normally much easier to deal with breaking ones, in only because the set of upgrades since the last succesful job is then explicit from a diff.

neutrinoceros:
close #508

### DIFF
--- a/.github/bottleneck-action/action.yml
+++ b/.github/bottleneck-action/action.yml
@@ -4,15 +4,16 @@ runs:
   using: composite
   steps:
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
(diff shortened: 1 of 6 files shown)


rgommers (approved):
Thanks! I reviewed those CI logs and the links in the lock file as best as I can, all seems to be in order. So let's give it a go and see if it helps make CI more stable.

PR #546, opened 2026-04-03
CLN: drop unused include dir

neutrinoceros:
This bit me in #545.
@rgommers suggested dropping the directory entirely, which seems like the right call to me.


rgommers (approved):
Thanks!

PR #565, opened 2026-05-31
TST: add a custom pytest session header to make runtime state (free-threading, JIT) visible

neutrinoceros:
`runtime-introspect` is a tiny library authored by yours truly. It's also distributed on conda-forge.

### DIFF
--- a/bottleneck/conftest.py
+++ b/bottleneck/conftest.py
@@ -1,8 +1,10 @@
+import textwrap
 from importlib.util import find_spec
 
 import pytest
 
 HAS_PYTEST_PARALLEL_RUN = find_spec("pytest_run_parallel") is not None
+HAS_RUNTIME_INTROSPECT = find_spec("runtime_introspect") is not None
 
 
 def pytest_configure(config):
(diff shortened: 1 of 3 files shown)


neutrinoceros:
to my surprise, the header still doesn't appear here even after #566. I'm wondering if an embedded `conftest.py` is just not collected in time for this hook to have a visible effect.

neutrinoceros:
Indeed, per [the documentation](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_report_header)
> This hook is only called for [initial conftests](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#pluginorder).

though from what I grasp, `bottleneck/conftest.py` should still be picked up as an "initial" conftest, as it lives in the parent directory to `bottleneck/tests`. It's possible the phrasing is just misleading, or that the discovery logic is actually broken in pytest itself.

neutrinoceros:
https://github.com/pytest-dev/pytest/issues/14557

neutrinoceros:
#573 makes this work as expected for `pytest .`, but not with `--pyargs`. I don't think CI should *only* exercise this mode, and I'd be in favor of changing our primary way to run tests to the default pytest mode.

PR #569, opened 2026-06-02
skip length-one axes when choosing flat stride in init_iter_all

### LINKED ISSUE #381: [BUG] Segfault or wrong value on array with non-vanilla strides.

### LINKED ISSUE #393: [BUG] Segmentation fault when working on a transposed numpy array with first dimension 1.
Issue raised as follow-up to: https://github.com/pydata/xarray/issues/6002 

**Reproduce**

```python
import numpy as np
import bottleneck as bn

n_time = 1
spec_data = np.random.random(size=(n_time,192,121))

bn.nanmax(spec_data.transpose(0, 2, 1))


 ---> Segmentation fault
```

numpy.transpose returns a view, so I guess that's what causes bottleneck to segfault? Not sure, though, especially since changing the order does not trigger the segfault: spec_data.transpose(1, 0, 2)... maybe bottleneck doesn't like views with a first dimension of size 1?

**Expected behaviour**
Should not crash

**Environment**
Confirmed Windows & Linux , P37 and P39. 

confirmed with:
bottleneck: 1.3.2
numpy: 1.21.4

sahvx655-wq:
When a reduce-all function gets a transposed view whose size-1 axis carries a large stride (`np.zeros((1,500,2)).transpose(1,2,0)`, `np.zeros((1,192,121)).transpose(0,2,1)`), the stride-picking loops in init_iter_all skip only zero strides and grab that axis's huge stride, so iteration runs off the buffer (issues #381, #393: garbage from nanmax or a segfault). Skip length-one axes there too, since their stride is never traversed.

### DIFF
--- a/bottleneck/tests/reduce_test.py
+++ b/bottleneck/tests/reduce_test.py
@@ -212,7 +212,9 @@ def test_nanstd_issue60():
         # issue #381 (C-contiguous branch)
         np.zeros((1, 500, 2)).transpose(1, 2, 0),
         # issue #393 (F-contiguous branch)
-        np.arange(1 * 192 * 121, dtype=np.float64).reshape(1, 192, 121).transpose(0, 2, 1),
+        np.arange(1 * 192 * 121, dtype=np.float64)
+        .reshape(1, 192, 121)
+        .transpose(0, 2, 1),
     ),
     ids=("C-contiguous", "F-contiguous"),
 )
--- a/bottleneck/tests/reduce_test.py
+++ b/bottleneck/tests/reduce_test.py
@@ -206,6 +206,25 @@ def test_nanstd_issue60():
     assert_equal(f, s, err_msg="issue #60 regression")
 
 
+def test_reduce_all_size_one_stride():
+    """reduce-all regression test (issues #381, #393)
+
+    A transposed view whose size-1 axis carries a large stride used to make
+    init_iter_all pick that stride as the flat element stride and iterate off
+    the buffer (garbage on the C-contiguous branch, segfault on the
+    F-contiguous branch).
+    """
+
+    # issue #381 (C-contiguous branch)
+    a = np.zeros((1, 500, 2)).transpose(1, 2, 0)
+    assert_equal(bn.nanmax(a), bn.slow.nanmax(a))
+
+    # issue #393 (F-contiguous branch)
+    b = np.arange(1 * 192 * 121, dtype=np.float64).reshape(1, 192, 121)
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Nice. Thank you for addressing these. Can you add regression tests too ? The example reproducers from issues you're fixing are probably good enough already.

sahvx655-wq:
Added a regression test in reduce_test.py covering both reproducers (#381 C-contiguous, #393 F-contiguous), checking bn.nanmax against bn.slow.nanmax on the transposed views.

neutrinoceros:
Great. Though these are really two different test cases. I suggest splitting them or, if it makes sense, to parametrize the one test function.

sahvx655-wq:
Parametrized it into the two cases (issue381, issue393).

bottleneck/tests/reduce_test.py
```diff
@@ -206,6 +206,27 @@ def test_nanstd_issue60():
...
 
+@pytest.mark.parametrize(
+    "a",
+    (
+        # issue #381 (C-contiguous branch)
+        np.zeros((1, 500, 2)).transpose(1, 2, 0),
+        # issue #393 (F-contiguous branch)
+        np.arange(1 * 192 * 121, dtype=np.float64).reshape(1, 192, 121).transpose(0, 2, 1),
+    ),
+    ids=("issue381", "issue393"),
```
neutrinoceros:
```suggestion
    ids=("C-contiguous", "F-contiguous"),
```

sahvx655-wq: Done, switched the ids to C-contiguous/F-contiguous.

neutrinoceros (approved):

neutrinoceros:
Thanks a lot for triaging and fixing these @sahvx655-wq !

PR #571, opened 2026-06-03
free output array and median heap on window==1 in move_median

sahvx655-wq:
move_median allocates its output array with PyArray_EMPTY in INIT, then both dtype variants take an early return when window is 1. The float variant frees the median heap there but leaks that output array; the integer variant leaks both the array and the heap because it never calls mm_free either. I noticed it reading the two siblings side by side. Repeated calls with window=1 leak an input-sized array every time, so it accumulates fast. Free the array and the heap before returning, matching the normal exit path.


neutrinoceros (approved):
LGTM. Thank you !

PR #589, opened 2026-09-11
TST/WHL: add support for CPython 3.15

neutrinoceros:


neutrinoceros:
blocked by #590

pyproject.toml
```diff
@@ -89,6 +90,9 @@ versionfile_build = "bottleneck/_version.py"
 tag_prefix = "v"
 parentdir_prefix = "bottleneck-"
 
+[tool.uv]
+no-build-package = ["numpy"]
```
rgommers, line 94: Preventative, or was `numpy` missing a wheel on some platform?

neutrinoceros: preventative
but that's also how I was reminded that 3.13t wasn't trivial to lock for, hence #590

Issue #591, opened 2026-09-12
REL: new release needed


neutrinoceros:
A new release is needed in order to publish wheels for CPython 3.15.
I already merged #589 which added it to our build matrix.
I also want to note that #511 would help make this a non-problem for GIL-enabled builds as well as vastly reduce the number of artifacts we need to publish to support all current versions of CPython as well as future ones. I also plan to explore the recently standardised `abi3t` to also fix this yearly point of friction for free-threaded builds, though the ecosystem (setuptools in particular) isn't ready yet.
If for any reason we *didn't* include #511 in our next release, I would also love feedback on wether we could/should drop support for 3.10, which will be EOL in a little less than a month. I don't know of a motivation to do so right now, other than reducing the number of wheels, which #511 would be much more effecient at.

neutrinoceros:
ping @rgommers @rdbisme

rgommers:
Agreed, let's ship a release. It'd be nice to include #511 - doing that before rather than after the next release will give us quick signal in case anything turned out to be wrong.

We can drop 3.10 now or right after, it doesn't matter to me. NumPy already dropped 3.11 so it's kinda pointless to keep shipping 3.10 wheels.

rdbisme:
Ok for me and indeed let's include #511

PR #594, opened 2026-09-12
DEP: drop support for CPython 3.10, require 3.11 or newer

neutrinoceros:
See #591
I have no strong opinion about this drop, which I think is unnecessary, possibly irrelevant with the work I'm doing with abi3 support, but also possibly nice-to-have if we're ever going to introduce type-checking to bottleneck, as 3.11 has nice features in this area I might like to use.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.10"
+          - "3.11"
           - "3.15"
         os:
           - ubuntu-22.04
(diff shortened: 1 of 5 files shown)


rdbisme:
Ok for me to deprecate 3.10

neutrinoceros:
ok then 3/3, let's merge this one now, if only to resolve merge conflicts early. Thanks !

rgommers:
> possibly irrelevant with the work I'm doing with abi3 support,

I think it helps quite a bit. Python 3.11 and 3.12 both had significant Limited C API improvements, some projects consider 3.11 the first usable version, some 3.12 (because vectorcall support landed in 3.12). After 3.12 it's been quite stable, only minor additions that few projects need.

PR #595, opened 2026-09-12
DEP: drop support for numpy 1.x

neutrinoceros:
Originally suggested by @rgommers in https://github.com/pydata/bottleneck/pull/556#pullrequestreview-5186485782 to help simplify the build process with meson.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,17 +3,8 @@ requires = [
     "setuptools>=77.0.1",
     "versioneer>=0.29",
     "versioneer[toml] ; python_version < '3.11'",
-    # Comments on numpy build requirement range:
-    #
-    #   1. >=2.0.x is the numpy requirement for wheel builds for distribution
-    #      on PyPI - building against 2.x yields wheels that are also
-    #      ABI-compatible with numpy 1.x at runtime.
-    #   2. Note that building against numpy 1.x works fine too - users and
-    #      redistributors can do this by installing the numpy version they like
-    #      and disabling build isolation.
-    #   3. The <2.(N+3) upper bound is for matching the numpy deprecation policy,
-    #      it should not be loosened more than that.
-    "numpy>=2,<2.6"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -3,16 +3,8 @@ requires = [
...
-    #   1. >=2.0.x is the numpy requirement for wheel builds for distribution
-    #      on PyPI - building against 2.x yields wheels that are also
-    #      ABI-compatible with numpy 1.x at runtime.
-    #   2. Note that building against numpy 1.x works fine too - users and
-    #      redistributors can do this by installing the numpy version they like
-    #      and disabling build isolation.
-    #   3. The <2.(N+3) upper bound is for matching the numpy deprecation policy,
-    #      it should not be loosened more than that.
+    # The <2.(N+3) upper bound is for matching the numpy deprecation policy,
+    # it should not be loosened more than that.
```
neutrinoceros: ironically, it is *already* looser than described. I don't see a problem with keeping the bounds we have now, so maybe I should just adapt the comment ?

rdbisme: Yep, sounds good.

pyproject.toml
```diff
@@ -3,16 +3,10 @@ requires = [
...
-    #   2. Note that building against numpy 1.x works fine too - users and
-    #      redistributors can do this by installing the numpy version they like
-    #      and disabling build isolation.
-    #   3. The <2.(N+3) upper bound is for matching the numpy deprecation policy,
-    #      it should not be loosened more than that.
+    # The upper bound should be at least 3 minor releases ahead of the lower bound
+    # in order to match numpy's deprecation policy.
+    # Anything beyond that needs to be tested against in CI, with deprecation warnings
+    # treated as errors.
     "numpy>=2,<2.6"
```
rgommers: Please bump to `<2.8`, since we're already at 2.5 now.

rgommers (approved):
This LGTM, thanks!

PR #596, opened 2026-09-12
RFC: explicitly cast arguments to `Py_INCREF` and `Py_DECREF`

neutrinoceros:
This resolves a bunch of compile-time warnings that are treated as errors on the meson branch (#556).


neutrinoceros:
Whoops. I wasn't expecting a segfault here. I'll try to keep this minimal

